Issue #7, opened 2022-10-05
Create a unified API between PPO and ILQL


LouisCastricato:
Currently ILQL was implemented more or less in a vacuum of our PPO implementation. As such, ILQL has features that our PPO implementation needs. This includes

- Removing the dependency on GPT2 for PPO. This is currently being done [here](https://github.com/samp830/trlx), but we can create a more standardized approach
- Adding the ability to pass custom reward functions to the orchestrator. This is an absolute must have.
- Standardized naming conventions between functions.

LouisCastricato:
@reciprocated @Dahoas no rush on this but whenever you guys get a chance we should discuss this in this issue.

LouisCastricato:
Fixed the first point with #11 

All that remains is standardizing naming conventions.

reciprocated:
On the topic of unified api, to encourage new users I vote for simplifying down to

```python
import trlx
# for ppo
trlx.train(model_path, reward_fn)
# for ilql
trlx.train(model_path, samples, ratings)
```

LouisCastricato:
I like that, good idea

PR #8, opened 2022-10-05
Adds style file and reward function capabilities to ppo orchestrator 

LouisCastricato:
This is in relation to #7 , we added a style file and code of conduct to make contributions easier. Similarly, we added the ability to pass custom reward functions to orchestrators. Note that you can still override the reward function within an orchestrator, but the hope is that this will not be the default going forward.


LouisCastricato:
@Dahoas, we need to update the colab notebook.

reciprocated (approved):
lgtm

Issue #29, opened 2022-10-11
[bug] Support for Soft Prompts in PPO Model


daia99:
Towards replication of ELM Stage 3, I'm looking into adding softprompts to train a conditional learnable embedding with PPO for each terrain mentioned in the paper.

Following https://github.com/kipgparker/soft-prompt-tuning.

Outlined code snippet example, and tracebacks for varying number of softprompt tokens. Will come back to this, but let me know if you have any suggestions for modifying the orchestrator.

Using the ppo_sentiments example and soft prompt implementation:
```python
if __name__ == "__main__":
    cfg = TRLConfig.load_yaml("configs/ppo_config.yml")

    sentiment_pipe = pipeline(
        "sentiment-analysis", "lvwerra/distilbert-imdb", device=-1
    )

    def reward_fn(samples: List[str]):
        sent_kwargs = {
            "return_all_scores": True,
            "function_to_apply": None,
            "batch_size": cfg.method.chunk_size,
        }
        pipe_outputs = sentiment_pipe(samples, **sent_kwargs)
        scores = torch.tensor([output[1]["score"] for output in pipe_outputs])
        return scores

    model: AcceleratePPOModel = get_model(cfg.model.model_type)(cfg)

    # setup soft prompt embeddings with 'n' prefix tokens, init from model vocab
    n_tokens = 1
    initialize_from_vocab = True
    
    s_wte = SoftEmbedding(model.model.gpt.get_input_embeddings(), 
                      n_tokens=n_tokens, 
                      initialize_from_vocab=initialize_from_vocab)

    model.model.gpt.set_input_embeddings(s_wte)
    
    pipeline: PPOPipeline = get_pipeline(cfg.train.pipeline)(model.tokenizer, cfg)
    orch: PPOOrchestrator = get_orchestrator(cfg.train.orchestrator)(
        model, pipeline, reward_fn=reward_fn, chunk_size=cfg.method.chunk_size
    )
    orch.make_experience(cfg.method.num_rollouts)
    model.learn()

    print("DONE!")
```

When `n_tokens = 1`, the following error occurs:
```
Traceback (most recent call last):
  File "/home/aleph/adai/trlx/examples/ppo_sentiments.py", line 101, in <module>
    orch.make_experience(cfg.method.num_rollouts)
  File "/home/aleph/adai/trlx/trlx/orchestrator/ppo_orchestrator.py", line 69, in make_experience
    logits, _, v = self.rl_model.model(all_tokens)
[...]

PR #32, opened 2022-10-12
Support for Soft Prompts in PPO Model

daia99:
For adding support to soft prompts in models, fixes https://github.com/CarperAI/trlx/issues/29.

Changes:
- Sample config and example added with minor changes from original ppo versions.
- ModelConfig has two new params to support soft prompt settings.
- New accelerate model added with soft prompt implementation.

Not tested with full training run, but runs training steps in added example script.

### DIFF
--- a/examples/ppo_softprompt_sentiments.py
+++ b/examples/ppo_softprompt_sentiments.py
@@ -1,0 +1,45 @@
+from typing import List
+
+import torch
+from torch import nn
+from transformers import pipeline
+
+import wandb
+from trlx.data.configs import TRLConfig
+from trlx.model.accelerate_ppo_model import AcceleratePPOModel
+from trlx.model.accelerate_ppo_softprompt_model import AcceleratePPOSoftpromptModel
+from trlx.orchestrator.ppo_orchestrator import PPOOrchestrator
+from trlx.pipeline.ppo_pipeline import PPOPipeline
+from trlx.utils.loading import get_model, get_orchestrator, get_pipeline
+
+
+if __name__ == "__main__":
+    cfg = TRLConfig.load_yaml("configs/ppo_softprompt_config.yml")
+
+    sentiment_pipe = pipeline(
+        "sentiment-analysis", "lvwerra/distilbert-imdb", device=-1
+    )
+
+    def reward_fn(samples: List[str]):
+        sent_kwargs = {
+            "return_all_scores": True,
+            "function_to_apply": None,
+            "batch_size": cfg.method.chunk_size,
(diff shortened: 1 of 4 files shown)


trlx/data/configs.py
```diff
@@ -29,6 +29,8 @@ class ModelConfig:
     model_type : str # One of the architectures present in framework.model
     device : str = ''
     num_layers_unfrozen : int = -1
+    n_soft_tokens: int = None # soft prompt support
```
LouisCastricato, line 32: You should be making a new model config for this so it isnt inherited by other model configs except your own.

trlx/model/accelerate_ppo_softprompt_model.py
```diff
@@ -0,0 +1,138 @@
...
+    def __init__(self, config, train_mode=True):
+        self.store = PPORolloutStorage()
+        super().__init__(config, self.store)
+
+        assert config.model.n_soft_tokens > 0, "Number of soft prompt tokens should be >=1"
+        
+        self.soft_dummy_token_id = 50256 # dummy token for padding soft prompts
+        self.n_soft_tokens = config.model.n_soft_tokens
+
+        s_wte = SoftEmbedding(self.model.gpt.get_input_embeddings(),
```
LouisCastricato, line 88: A few comments about what this does would be great.

examples/ppo_softprompt_sentiments.py
```diff
@@ -0,0 +1,45 @@
...
+        sent_kwargs = {
+            "return_all_scores": True,
+            "function_to_apply": None,
+            "batch_size": cfg.method.chunk_size,
+        }
+        pipe_outputs = sentiment_pipe(samples, **sent_kwargs)
+        scores = torch.tensor([output[1]["score"] for output in pipe_outputs])
+        return scores
+
+    model: AcceleratePPOSoftpromptModel = get_model(cfg.model.model_type)(cfg)
```
LouisCastricato, line 33: You don't need this. Just remove the type case from ppo_sentiments and leave the imports. The registery (e.g. get_model) will be able to determine which model type to load based off the config you use.

LouisCastricato:
Needs work.

Dahoas:
Thanks for the pr! It just occurred to me that I'm not sure how training soft-prompts will play with the usual practice of freezing the bottom 2/3 transformer blocks of the model. Surely there's some kind of work-around but I haven't thought of anything at the moment. 

At any rate things should be fine for small models which are fine-tuned without freezing any layers.

Issue #41, opened 2022-10-18
Installation error due to multiple top-level packages 


jon-tow:
# Issue

Installing `trlx` in a fresh environment, following the README.md guide (`python setup.py develop`),  results in the following package error:

```bash
error: Multiple top-level packages discovered in a flat-layout: ['trlx', 'configs', 'unittests'].

To avoid accidental inclusion of unwanted files or directories,
setuptools will not proceed with this build.

If you are trying to create a single distribution with multiple packages
on purpose, you should not rely on automatic discovery.
Instead, consider the following options:

1. set up custom discovery (`find` directive with `include` or `exclude`)
2. use a `src-layout`
3. explicitly set `py_modules` or `packages` with a list of names

To find more information, look for "package discovery" on setuptools docs.
```

# Environment

```bash

Python version : 3.10.6

Compiler    : Clang 13.1.6 (clang-1316.0.21.2.5)
OS          : Darwin
Release     : 21.5.0
Machine     : arm64
Processor   : arm
```

jon-tow:
Addressed with #42

PR #42, opened 2022-10-18
[update] Improve package setup

jon-tow:
This PR addresses package setup issues with the following updates:

- Adds package finding to `setup.cfg` to avoid "multiple top-level packages" errors during new environment installs. 
  - __Related issue__: #41
- Updates `README.md` installation instructions to use `pip install -e ...`, as  `python setup.py ...` is deprecated and should be replaced with pip tools ([reference guide](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html)).
- Adds separate `dev` dependencies to `setup.cfg` to avoid forcing `dev` dependencies on package install.
    - Adopts `pytest` as the test runner.
- Renames `unittests` to `tests` to avoid confusion with `unittest` module and adhere to common Python unit-test directory naming.


Dahoas:
This is great thanks!

Issue #40, opened 2022-10-18
fix type errors and add mypy 


cat-state:
We should also add it as a pre-merge hook

jon-tow:
Note: #43 removes the `flake8` F82 undefined-name check. This needs to be caught by `mypy`.

PR #43, opened 2022-10-18
Add initial GitHub workflows

jon-tow:
This PR adds initial GitHub workflows for building, testing, and checking for code quality.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,6 +1,6 @@
 ---
 name: 🐛 Bug Report
-description: Report a bug or unexpected behavior to help us improve trlX 
+description: Report a bug or unexpected behavior to help us improve trlX
 labels:
 - bug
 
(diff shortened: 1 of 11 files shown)


jon-tow:
This relies on #42 for automatic `flake8` dependency installation in the `build.yml` run.

jon-tow:
I've removed the "undefined name" check in the build workflow. CC @cat-state if they think we should add `mypy` in this PR or open a new one when we figure out how to properly set it up.

Dahoas:
If we think `mypy` can be set up by end of the week I'm for adding to this PR. If not we can merge now.

LouisCastricato:
Is this good to merge?

cat-state:
I think let's do mypy on a subsequent one cus we'll have type errors and annotations to fix

jon-tow:
Le'ts merge. I'm currently unfamiliar with `mypy` and Aman might be too busy to add it right now. We can move discussions/concerns to issue #40.

PR #57, opened 2022-10-21
Fix grammar (tense)

mrm8488:


LouisCastricato:
Don't make a PR like this unless its a substantial number of changes. We will not merge a PR that only changes one word.

PR #76, opened 2022-10-28
Hyperparameter Optimization with Ray Tune and Weights and Biases

ayulockin:
# Hyperparameter optimization for TRLX:

## Stack:

* Ray Tune for distributed hyperparameter optimization.
* Weights and Biases for tracking and for finding the best hyperparameters.

## Context:

The hyperparameter optimization system is designed based on the structure of the `examples/` provided. For now I have built the system using `examples/ppo_sentiments.py`.

### Setup search space

In Ray Tune's literature we need to define a `param_space` which is passed to `tune.TuneConfig(...)`. To define a `param_space`:

* Copying the default config provided in `configs/*.yml` to `configs/ray_tune_configs/my_search_config.yml`.
* Use `strategy` and `values` to define the space as shown below:
    ````
    seq_length:  # Size of LM context
        strategy: "choice"
        values: [36, 48, 52]
    ````
    The allowed strategies are based on what's available by Ray Tune as per the docs [here](https://docs.ray.io/en/latest/tune/api_docs/search_space.html#tune-sample-docs). The `strategy` and `values` are parsed using `get_param_space` function. 

### Setup tune config

The tune config describes the search algorithm, scheduler, metric and mode. This is defined in the `configs/ray_tune_configs/my_search_config.yml` itself.

- Search Algo: Bayesian Optimization with Hyberband ([BOHB](https://docs.ray.io/en/latest/tune/api_docs/suggestion.html#bohb-tune-search-bohb-tunebohb)) is tested and working. Use `bohb` to enable it. For random search use `null`.
- Scheduler: BOHB requires `[HyperBandForBOHB](https://docs.ray.io/en/latest/tune/api_docs/schedulers.html#tune-scheduler-bohb)`; pass `hyperbandforbohb` to enable it.

### The training function

The training function takes a `config` argument. It's the `ppo_sentiments.py` example packaged as a function. Multiple training functions can be implemented using the same format.

### The `train_sweep.py` file

The main logic is implemented here. 

## Usage

`python train_sweep.py --config configs/ray_tune_configs/ppo_config.yml --example-name ppo_sentiments`

## Notable changes

* The W&B was initialized inside the `accelerate_base_model.py`; I have moved it to `trlx.py`.
* W&B logging is disabled when `ray.is_initialized`. Ray Tune initializes multiple concurrent trials (experiments), and W&B was erroring out for a few trials and then a new run was initialized with metics logged to the errored out run. It was not an issue with random or grid search but with bayesian search only.

## How is W&B tracking done?

Once all the trails are done, the metrics are logged using the `log_trials` function. Since we are not tracking live we will not have access to system metrics. But since we are looking to find the best hyperparameters and keep track of the experiments, I logged the metrics after the experiments are done. This also makes it flexible for any search algorithm to be used especially those that are hard to parallelize like Bayes opt.

## Result

The experiments are logged to the wandb `project_name`; example [here](https://wandb.ai/ayut/trlx-hyperopt-bohb?workspace=user-ayut). We can easily convert it to W&B sweep using the UI. 

https://user-images.githubusercontent.com/31141479/199493284-082d0955-c705-4518-8a9c-0a26916a8603.mov

### The most relevant hyperparameters

We can use the parameter importance chart.

image

We can easily find the best parameters from the parallel importance plot.

https://user-images.githubusercontent.com/31141479/199494075-600709aa-ebea-42a1-a023-573b7c465b89.mov

# TODOs:

- [ ] The example is running using a single GPU per trial. We can easily provision multiple GPUs for each trial but I am working to distribute the training. The recommended solution is to use Ray Train as suggested [here](https://docs.ray.io/en/latest/tune/tutorials/tune-resources.html#how-to-run-distributed-training-with-tune). 
- [ ] Automatically build W&B report with the analysis of the hyperparameter optimization job.

### DIFF
--- a/configs/ray_tune_configs/ppo_config.yml
+++ b/configs/ray_tune_configs/ppo_config.yml
@@ -26,12 +26,11 @@ train:
   total_steps: 10000  # Train for max(epochs, total_steps)
   batch_size: 128  # batch size
 
-  lr_ramp_steps: 100  # learning rate warm up
-  lr_decay_steps: 79000  # learning rate decay
-  weight_decay: 1.0e-6  # weight decay param
-  learning_rate_init: 1.412e-4  # init learning rate
-  learning_rate_target: 1.412e-4  # target final learning rate
+  lr_init: 1.412e-4  # init learning rate
+  lr_target: 1.412e-4  # target final learning rate
   opt_betas: [0.9, 0.95] # adam betas
+  opt_eps: 1.0e-8  # adam eps
+  weight_decay: 1.0e-6  # weight decay param
 
   checkpoint_interval: 10000  # checkpoint interval
   eval_interval: 4  # eval interval
--- a/configs/ray_tune_configs/ppo_config.yml
+++ b/configs/ray_tune_configs/ppo_config.yml
@@ -14,12 +14,12 @@ model:
   model_type: "AcceleratePPOModel"  # Name of accelerate model type to load
   num_layers_unfrozen:  # Number of bottom layers to freeze during training
     strategy: "choice"
-    choices: [2, 3]
+    values: [2, 3]
 
 train:
   seq_length:  # Size of LM context
     strategy: "choice"
-    choices: [36, 48, 52]
+    values: [36, 48, 52]
   epochs: 1  # Train for max(epochs, total_steps)
   total_steps: 1000  # Train for max(epochs, total_steps)
   batch_size: 128  # batch size
@@ -42,20 +42,20 @@ method:
   name: 'ppoconfig'  # Name of RL method config
   num_rollouts:  # Number of rollouts to collect per epoch
     strategy: "choice"
-    choices: [96, 128]
+    values: [96, 128]
   chunk_size: 128  # Number of rollouts to collect in one loop of orchestrator
   ppo_epochs:  # Number of ppo epochs
     strategy: "randint"
-    bounds: [3, 6] # 3 is inclusive, 6 is exclusive
+    values: [3, 6] # 3 is inclusive, 6 is exclusive
   init_kl_coef: # init kl coefficient
     strategy: "quniform"
-    bounds: [0.1, 0.3, 0.1]
+    values: [0.1, 0.3, 0.1]
   target: 6  # target kl coefficient, set None for fixed kl coef
   horizon: 10000  # PPO horizon
   gamma: 1  # PPO discount
   lam: # PPO lambda
     strategy: "uniform"
-    bounds: [0.93, 0.98]
+    values: [0.93, 0.98]
   cliprange: 0.2  # clip range
   cliprange_value: 0.2  # clip range
   vf_coef: 2.3  # value term weight
--- a/configs/ray_tune_configs/ppo_config.yml
+++ b/configs/ray_tune_configs/ppo_config.yml
@@ -3,7 +3,7 @@ tune_config:
   metric: "mean_reward"
   search_alg: "bohb" # null
   scheduler: "hyperbandforbohb" # null
-  num_samples: 1
+  num_samples: 15
   max_concurrent_trials: null
   time_budget_s: null
   reuse_actors: null
@@ -20,8 +20,10 @@ train:
   seq_length:  # Size of LM context
     strategy: "choice"
     values: [36, 48, 52]
-  epochs: 1  # Train for max(epochs, total_steps)
-  total_steps: 100  # Train for max(epochs, total_steps)
+  epochs:  # Train for max(epochs, total_steps)
+    strategy: "choice"
+    values: [80, 100, 120]
+  total_steps: 10000  # Train for max(epochs, total_steps)
   batch_size: 128  # batch size
 
   lr_ramp_steps: 100  # learning rate warm up
@@ -32,11 +34,11 @@ train:
   opt_betas: [0.9, 0.95] # adam betas
 
   checkpoint_interval: 10000  # checkpoint interval
-  eval_interval: 2  # eval interval
+  eval_interval: 4  # eval interval
 
   pipeline: "PPOPipeline"  # prompt pipeline to load
   orchestrator: "PPOOrchestrator"  # orchestrator to load
-  project_name: "trlx-hyperopt-random"
+  project_name: "trlx-hyperopt-bohb"
 
 method:
   name: 'ppoconfig'  # Name of RL method config
--- a/configs/ray_tune_configs/ppo_config.yml
+++ b/configs/ray_tune_configs/ppo_config.yml
@@ -3,7 +3,7 @@ tune_config:
   metric: "mean_reward"
   search_alg: "bohb" # null
   scheduler: "hyperbandforbohb" # null
-  num_samples: 10
+  num_samples: 1
   max_concurrent_trials: null
   time_budget_s: null
   reuse_actors: null
--- a/configs/ray_tune_configs/ppo_config.yml
+++ b/configs/ray_tune_configs/ppo_config.yml
@@ -3,7 +3,7 @@ tune_config:
   metric: "mean_reward"
   search_alg: "bohb" # null
   scheduler: "hyperbandforbohb" # null
-  num_samples: 1
+  num_samples: 10
   max_concurrent_trials: null
   time_budget_s: null
   reuse_actors: null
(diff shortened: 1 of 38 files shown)


trlx/trlx.py
```diff
@@ -90,4 +118,8 @@ def train(
         raise ValueError(f"Either {dataset=} or {reward_fn=} should be given")
 
     model.learn()
+
+    if accelerator.is_main_process and not ray.is_initialized():
+        accelerator.end_training()
```
reciprocated, line 123: this is equivalent to `run.finish()` if you were to reuse that variable, but is this also a required thing since it was not present before.

trlx/ray_tune/wandb.py
```diff
@@ -0,0 +1,60 @@
...
+import wandb
+import json
+import pandas as pd
+from pathlib import Path
+
+
+def parse_result(result):
+    tmp_result = {}
+    for key, value in result.items():
+        if not "time" in key and isinstance(value, (int, float)):
```
reciprocated, line 13: this rather specific filter, since these timing variables might also be of interest. can this be changed to something less strict perhaps?

train_sweep.py
```diff
@@ -0,0 +1,88 @@
...
+        "--example-name", type=str, default="ppo_sentiments", help="Name of the example"
+    )
+    parser.add_argument(
+        "--config", type=str, default=None, required=True, help="The config file defining the param_space."
+    )
+    parser.add_argument(
+        "--num-cpus", type=int, default=4, help="Number of CPUs to use per exp."
+    )
+    parser.add_argument(
+        "--num-gpus", type=int, default=0, help="Number of GPUs to use per exp."
```
reciprocated, line 53: default=1

train_sweep.py
```diff
@@ -0,0 +1,88 @@
...
+    )
+    parser.add_argument(
+        "--server-address",
+        type=str,
+        default=None,
+        required=False,
+        help="The address of server to connect to if using Ray Client.",
+    )
+
+    args, _ = parser.parse_known_args()
```
reciprocated, line 63: maybe `parse_args()` to remove confusion

configs/ray_tune_configs/ppo_config.yml
```diff
@@ -0,0 +1,69 @@
+tune_config:
+  mode: "max"
+  metric: "mean_reward"
+  search_alg: "bohb" # null
```
reciprocated, line 4: maybe it's worth making random search option explicit instead of using `null` for it

configs/ray_tune_configs/ppo_config.yml
```diff
@@ -0,0 +1,69 @@
...
+  scheduler: "hyperbandforbohb" # null
```
reciprocated, line 5: same for fifo here

reciprocated (commented):
Overall this looks great, thank you for the effort!

I have a few more general questions about these changes.

> The W&B was initialized inside the accelerate_base_model.py; I have moved it to trlx.py.

Can you explain why this had to be done?

> W&B logging is disabled when ray.is_initialized. Ray Tune initializes multiple concurrent trials (experiments), and W&B was erroring out for a few trials and then a new run was initialized with metics logged to the errored out run. It was not an issue with random or grid search but with bayesian search only.

It seems to me hyperband as is, is a bit crippled, or maybe any other trial scheduler as well, since it keeps "pausing" runs, which effectively terminates them and then later restarts those runs from scratch. Have you looked into some way of customizing this process with resuming from checkpoints? Also can something similar may be done for wandb state?

Can you also help me with how to select parameter space for sweeps in W&B, since it doesn't pick up the correct ones. I follow your steps from the videos, but I still have to manually change sweep configuration. Had you done something special between your first video and the second one?

ayulockin:
Thank you for the review. 

> Can you also help me with how to select parameter space for sweeps in W&B, since it doesn't pick up the correct ones. I follow your steps from the videos, but I still have to manually change sweep configuration. Had you done something special between your first video and the second one?

Hey @reciprocated, to answer this quickly, I am adding a feature that will generate a W&B report with relevant charts so that you or any user will not have to worry about doing it manually. I am working on it.

trlx/ray_tune/__init__.py
```diff
@@ -0,0 +1,192 @@
...
+            assert len(value["values"]) == 2
+            return tune.uniform(*value["values"])
+        elif strategy == "quniform":
+            assert isinstance(value["values"], list)
+            assert len(value["values"]) == 3
+            return tune.quniform(*value["values"])
+        elif strategy == "loguniform":
+            assert isinstance(value["values"], list)
+            assert len(value["values"]) == 3
+            return tune.loguniform(*value["values"])
```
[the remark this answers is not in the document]
ayulockin: as per the documentation it should accept 3 values: https://docs.ray.io/en/latest/tune/api_docs/search_space.html#tune-loguniform

`ray.tune.loguniform(lower: float, upper: float, base: float = 10)`

We can probably do with two if we are okay with base = 10 as a good default.

trlx/ray_tune/wandb.py
```diff
@@ -0,0 +1,60 @@
...
+            config=dict(params),
+            group=trial_path.stem,
+            job_type="hyperopt",
+        )
+
+        # Open result.json and log the metrics to W&B.
+        with open(os.path.join(trial, "result.json"), "r") as f:
+            for line in f:
+                result = dict(json.loads(line))
+                result.pop("config", None)
```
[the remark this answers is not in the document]
ayulockin: since we have already logged the config from the `params.json` file we don't need this info.

trlx/trlx.py
```diff
@@ -36,13 +40,37 @@ def train(
...
+            project_name=config.train.project_name,
+            config=config.to_dict(),
+            init_kwargs={
+                "wandb": {
+                    "mode": "disabled" if os.environ.get("debug", False) else "online",
+                }
+            },
+        )
+
+        run = accelerator.get_tracker("wandb")
```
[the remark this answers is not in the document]
ayulockin: No, I might have missed it here while experimenting. 

We might use it though in the future but for now we can remove it.

trlx/model/accelerate_base_model.py
```diff
@@ -18,17 +17,28 @@
...
     from tqdm import tqdm
 
+import ray
+from ray.air import session
+
+
+def parse_results_for_session(results: dict):
+    for k, v in results.items():
+        if isinstance(v, torch.Tensor):
+            results[k] = float(v)
```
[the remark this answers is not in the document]
ayulockin: The `session.report(results)` write the metrics as a JSON file. If the value is `torch.Tensor` the metric is written as a string in the JSON file. We can get away with it but then will have to get the float value from the string. To simply it I implemented this feature.

trlx/model/accelerate_base_model.py
```diff
@@ -174,7 +169,7 @@ def evaluate(self):
                 columns.append("reward")
                 columns_data.append(rewards)
                 stats["mean_reward"] = mean_reward
-                print(f"{mean_reward=}")
+                # print(f"{mean_reward=}")
```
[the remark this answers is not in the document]
ayulockin: Sure.

trlx/model/accelerate_base_model.py
```diff
@@ -194,9 +189,8 @@ def evaluate(self):
                     columns_data.append(values)
 
             rows = list(zip(*columns_data))
-            stats["samples"] = wandb.Table(columns=columns, rows=rows)
-
-            print(rows[0])
+            if not ray.is_initialized():
+                stats["samples"] = wandb.Table(columns=columns, rows=rows)
```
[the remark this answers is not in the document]
ayulockin: The table construction is not working well with Ray Tune. Do we absolutely need it?

trlx/trlx.py
ayulockin: we can probably remove the `run` variable from above.

trlx/ray_tune/wandb.py
ayulockin: yes we can do that. Do you have a suggestion here.

configs/ray_tune_configs/ppo_config.yml
ayulockin: yeah makes sense. will do it.

trlx/ray_tune/__init__.py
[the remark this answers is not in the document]
reciprocated: Yeah, you're right. I very was thrown off by that assert

ayulockin:
> Hey @reciprocated, to answer this quickly, I am adding a feature that will generate a W&B report with relevant charts so that you or any user will not have to worry about doing it manually. I am working on it.

Hey @reciprocated, my latest commit adds the ability to create a W&B report automatically. I will share the result once the training is done.

LouisCastricato (approved):

Issue #53, opened 2022-10-20
PPO Implementation Details - Checklist


herbiebradley:
[The 37 Implementation Details of PPO](https://iclr-blog-track.github.io/2022/03/25/ppo-implementation-details), a blog post published at ICLR, details a number of PPO implementation details to improve both efficiency and model performance. See also: [Andrychowicz et al.](https://arxiv.org/abs/2006.05990), [Engstrom et al.](https://arxiv.org/abs/2005.12729)

Some of these optimizations are minor and probably irrelevant, many are already implemented here, and some may provide performance boosts to trlx. This issue documents these details as a checklist, to track the progress of this repository towards the entire list.

- [x] 1.  Vectorized Architecture  - `trlx` already does this.
- [ ] 2. Weights and Biases Initialisation. Any layers initialised from scratch should use orthogonal initialization with scaling `sqrt(2)` and bias of 0, with policy network last layer scaled by `0.01` after init.
- [x] 3. Adam Optimizer initialization. Andrychowicz et al. recommend `1e-7` as Adam epsilon (and actually find that the PyTorch default of `1e-8` is the worst of the choices tested).
- [x] 4. Optimizer Weight Decay. Currently the code does not use the config value of `weight_decay: 1e-6` at all? It also uses Cosine Annealing instead of Linear, and decays not to 0 (recommended by Andrychowicz et al.) but to `1.412e-4` by default. Maybe test linear to see if it makes a difference?
- [x] 5. Generalized Advantage Estimation. Correctly implemented in `trlx`.
- [x] 6. Mini-batch updates. In `trlx` this is being done in `make_experience`.
- [x] 7. Normalization of Advantages (at the mini-batch level). I believe this is being done, since I think `whiten` is called at mini-batch level?
- [x] 8. Clipped surrogate objective. Done in `trlx`.
- [x] 9. Value function loss clipping. Done in `trlx`.
- [ ] 10. Overall loss and entropy bonus. Entropy is not used for regularization in `trlx`. OAI set it to 0 for mujoco anyway, and Andrychowicz et al. find that regularization does not help performance, so this may not be useful to implement.
- [ ] 11. Global gradient clipping. The `trlx` `grad_clip` config option does not appear to be connected to anything. Andrychowicz et al. find a small performance boost from ensuring the norm of gradients of all parameters does not exceed `0.5`.
- [ ] 12. KL approximation. Check that the [unbiased estimator](http://joschu.net/blog/kl-approx.html) is being used.
[...]

PR #82, opened 2022-11-06
Update `TrainConfig` optimizer hyperparameters

jon-tow:
This PR provides the following updates based on observations from issue #53:
- Adds the previously unused `weight_decay` TrainConfig parameter to `AccelerateRLModel` optimizer initialization.
- Adds an `opt_eps` TrainConfig parameter to the `AccelerateRLModel` to allow users to specify an optimizer's epsilon value.
- Renames `learning_rate_init` and `learning_rate_target` to `lr_init` and `lr_target` TrainConfigs, respectively, for consistency with learning rate naming conventions in `torch.optim` algorithms.
- Removes unused `lr_ramp_steps` and `lr_decay_steps` TrainConfig parameters which look to be leftovers from the [magicCARP lr scheduler](https://github.com/EleutherAI/magiCARP/blob/777a578abcf0f790b653d3bce9dcbff8f5cd6bd7/carp/util.py#L25).

### DIFF
--- a/configs/ilql_config.yml
+++ b/configs/ilql_config.yml
@@ -10,12 +10,11 @@ train:
   epochs: 10
   total_steps: 10000
 
-  lr_ramp_steps: 100
-  lr_decay_steps: 3366
-  weight_decay: 1e-6
-  learning_rate_init: 1e-4
-  learning_rate_target: 1e-4
+  lr_init: 1.0e-4
+  lr_target: 1.0e-4
   opt_betas: [0.9, 0.95]
+  opt_eps: 1.0e-8
+  weight_decay: 1.0e-6
 
   checkpoint_interval: 1000
   eval_interval: 16
--- a/configs/ppo_config.yml
+++ b/configs/ppo_config.yml
@@ -10,12 +10,11 @@ train:
   total_steps: 10000  # Train for max(epochs, total_steps)
   batch_size: 128  # batch size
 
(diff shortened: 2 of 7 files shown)


LouisCastricato:
@herbiebradley since this is for your issue.

herbiebradley (approved):
Looks good! Can we change the default epsilon to `1.0e-7` (default in Tensorflow) since the ablation paper found it was significantly better?

Also, does anyone know where the weight decay of `1.0e-6` comes from, since it is significantly lower than the default?

reciprocated:
> Also, does anyone know where the weight decay of `1.0e-6` comes from, since it is significantly lower than the default?

From here https://github.com/EleutherAI/magiCARP/blob/3537b4d7c98a4a384964856740a65731746e04fe/configs/base_config.yml#L18

reciprocated (commented):
lgtm!

LouisCastricato (approved):

jon-tow:
@herbiebradley I intentionally left the default $\epsilon$ to `1.0e-8` as I wasn't 100% sure that the improvement would directly carry over to AdamW (we do not use vanilla Adam). I'd like to try some experiments before updating in another PR if that's okay.

jon-tow:
cc #76 for awareness

ayulockin:
Thanks for tagging the issue @jon-tow.

PR #89, opened 2022-11-11
ILQL loss refactor

cat-state:
In this, we refactor out the ILQL loss function and model additions so they can be reused with other accelerator libraries. 
I also refactored the loss to be slightly clearer and fixed some type errors.
First part of https://github.com/CarperAI/trlx/pull/75

W&B run: https://wandb.ai/carperai/trlx/runs/3tam2www


LouisCastricato:
LGTM. We should have @shahbuland update the readthedocs though.

LouisCastricato (approved):

cat-state:
> LGTM. We should have @shahbuland update the readthedocs though.

Maybe we should make the CI update them automatically on merge?

PR #90, opened 2022-11-11
Increase default eval interval

cat-state:
Much faster training - ~20 min instead of >1 hour. This is in units of train_steps
https://wandb.ai/carperai/trlx/runs/3gkbitl7


cat-state:
Actually turns out 16 is better, the progress bar is wrong instead.

cat-state:
Re-opening because increasing it still useful

LouisCastricato (approved):

PR #98, opened 2022-11-17
Add optional normalization (cont.)

reciprocated:
- change scaling reward to have three options: ref/running/none
- add single thread test of running mean calculations

https://wandb.ai/sorry/public/runs/2x8totg2

PR #99, opened 2022-11-18
Fix github build 

reciprocated:
- rename master to main


reciprocated:
merged along test fixes into #98

PR #100, opened 2022-11-18
Force class registry via imports

jon-tow:
This PR fixes the issue in which registered entities of `trlx` are not accessible by loading getters unless manually imported from higher-level calling sites by importing all such entities from `trlx/utils/loading.py`.

Updates:
- Fixes `ilql_config.yml` to use the proper `AccelerateILQLModel` type.

Notes:
- There isn't a distinction between `PPOPipeline` and `OfflinePipeline` anymore so I'm not sure if it's worth keeping that configuration option around.

### DIFF
--- a/configs/ilql_config.yml
+++ b/configs/ilql_config.yml
@@ -1,7 +1,7 @@
 model:
   model_path: "gpt2"
   tokenizer_path: "gpt2"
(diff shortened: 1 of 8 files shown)


LouisCastricato (approved):
LGTM

PR #108, opened 2022-11-22
Fix ppo ratio inaccuracy

### LINKED ISSUE #107: Ratio != 1 at start of PPO training (during loss function calculation)
### 🐛 Describe the bug

I've run ppo_sentiments.py, and an older version, and seeing that ratio is != 1 at step 0 (before optimizer step), at this line:
https://github.com/CarperAI/trlx/blob/main/trlx/model/nn/ppo_models.py#L165

https://iclr-blog-track.github.io/2022/03/25/ppo-implementation-details/ - reference regarding ratio = 1 at first epoch/mini-batch update:

> Check if ratio=1: Check if the ratio are always 1s during the first epoch and first mini-batch update, when new and old policies are the same and therefore the ratio are 1s and has nothing to clip. If ratio are not 1s, it means there is a bug and the program has not reconstructed the probability distributions used in rollouts.

When making experience, the ratio you'd get here at the start of training (before optimization) is 1: https://github.com/CarperAI/trlx/blob/main/trlx/orchestrator/ppo_orchestrator.py#L130

Seems like a currently unknown cause/bug, which leads to unexpected ratio values.

0.3.0

Python 3.8

reciprocated:
This pr fixes #107 
- lets models in orchestrator use `attention_mask` for computing `old_logprobs`

but also
- removes entity_name from ppo_config (which I suspect wasn't intentional)
- enables more thorough logging, synced with orchestrator's logging with `iter_count`
- adds git commit as tag to wandb runs

https://wandb.ai/sorry/public/reports/Ratio-fix--VmlldzozMDE4NzI2

### DIFF
--- a/configs/ppo_config.yml
+++ b/configs/ppo_config.yml
@@ -21,7 +21,6 @@ train:
 
   pipeline: "PromptPipeline"  # prompt pipeline to load
   orchestrator: "PPOOrchestrator"  # orchestrator to load
-  entity_name: "jon-tow"
 
 method:
   name: 'ppoconfig'  # Name of RL method config
--- a/trlx/model/accelerate_base_model.py
+++ b/trlx/model/accelerate_base_model.py
@@ -23,7 +23,7 @@ else:
 import ray
 from ray.air import session
 from ray.air.checkpoint import Checkpoint
-from trlx.utils import filter_non_scalars
+from trlx.utils import filter_non_scalars, get_git_tag
 
 
 @register_model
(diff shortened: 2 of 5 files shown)


Dahoas:
Perhaps add the ratio == 1 check in a unittest?

configs/ppo_config.yml
```diff
@@ -21,7 +21,6 @@ train:
 
   pipeline: "PromptPipeline"  # prompt pipeline to load
   orchestrator: "PPOOrchestrator"  # orchestrator to load
-  entity_name: "jon-tow"
```
jon-tow, line 24: Not sure how this made it through. Good catch and sorry!

jon-tow (approved):
Looks great @reciprocated 👏  Should we cut a patch release for this `0.3.1`?

Issue #106, opened 2022-11-21
randomwalks.py doesn't work on v0.3 or current main


dpaleka:
### 🐛 Describe the bug

On both `v0.3` and `https://github.com/CarperAI/trlx/commit/ff0d0776ce9189c7e0ebc954dd14bbca1136a450`, following the instructions from `README.md` and running
```
wandb disable && python examples/randomwalks.py
```
produces the following error:

```
Traceback (most recent call last):
  File "/home/dpaleka/code/trlx/examples/randomwalks.py", line 103, in <module>
    trlx.train(
  File "/home/dpaleka/code/trlx/trlx/trlx.py", line 95, in train
    model.learn()
  File "/home/dpaleka/code/trlx/trlx/model/accelerate_base_model.py", line 240, in learn
    results = self.evaluate()
  File "/home/dpaleka/code/trlx/trlx/model/accelerate_base_model.py", line 160, in evaluate
    samples = self.generate(prompts)
  File "/home/dpaleka/code/trlx/trlx/model/accelerate_base_model.py", line 133, in generate
    return self.accelerator.unwrap_model(self.model).generate(
  File "/home/dpaleka/code/trlx/trlx/model/nn/ilql_models.py", line 306, in generate
    logits[torch.where(logit_mask[input_ids[:, -1].squeeze()])] = -np.inf
RuntimeError: indices should be either on cpu or on the same device as the indexed tensor (cpu)
```

### Which trlX version are you using?

trlx==0.3

LouisCastricato:
Resolved https://github.com/CarperAI/trlx/pull/105

PR #109, opened 2022-11-22
updating gptj-config

Dahoas:
Updating the gptj config and verifying gptj runs on 8 A100s with zero2

https://wandb.ai/dahoas/trlx/runs/5vc7xsx8?workspace=user-dahoas

### DIFF
--- a/examples/ppo_sentiments.py
+++ b/examples/ppo_sentiments.py
@@ -17,7 +17,7 @@ def get_positive_score(scores):
     return dict(map(lambda x: tuple(x.values()), scores))["POSITIVE"]
 
 
-default_config = yaml.safe_load(open("configs/ppo_config.yml"))
+default_config = yaml.safe_load(open("configs/ppo_gptj.yml"))
 
 
 def main(hparams={}):
@@ -46,7 +46,6 @@ def main(hparams={}):
     prompts = [" ".join(review.split()[:4]) for review in imdb["text"]]
 
     model = trlx.train(
-        "lvwerra/gpt2-imdb",
         reward_fn=reward_fn,
         prompts=prompts,
         eval_prompts=["I don't know much about Hungarian underground"] * 64,
--- a/configs/ppo_gptj.yml
+++ b/configs/ppo_gptj.yml
@@ -35,6 +35,10 @@ method:
   cliprange: 0.2  # clip range
   cliprange_value: 0.2  # clip range
   vf_coef: 0.2  # value term weight
+  scale_reward: False # False | "ref" | "running" estimate against which to scale rewards
+  ref_mean: null
+  ref_std: null # rescale rewards with this deviation
+  cliprange_reward: 10
   gen_kwargs:
     max_length: 48  # LM max sample gen length
     min_length: 48  # LM min sample gen length


examples/ppo_sentiments.py
```diff
@@ -17,7 +17,7 @@ def get_positive_score(scores):
     return dict(map(lambda x: tuple(x.values()), scores))["POSITIVE"]
 
 
-default_config = yaml.safe_load(open("configs/ppo_config.yml"))
+default_config = yaml.safe_load(open("configs/ppo_gptj.yml"))
```
reciprocated, line 20: are you sure you want to change the default to gpt-j? I wouldn't mind the second script which imports `main` from here and changes config as the most simplest option right now, until we switch to hydra or something else

LouisCastricato: Agreed, this is a bad idea.

Dahoas: Yeah I just forgot to change this back.

Dahoas:
Added distributed config logging to wandb

LouisCastricato (approved):

configs/ppo_config.yml
```diff
@@ -21,6 +21,7 @@ train:
 
   pipeline: "PromptPipeline"  # prompt pipeline to load
   orchestrator: "PPOOrchestrator"  # orchestrator to load
+  entity_name: "dahoas"  # put your wandb login here
```
reciprocated, line 24: this will give somewhat cryptic error if you're not logged in as dahoas, maybe we can make this as optional enviroment variable instead?

LouisCastricato: Lets fix this in a future PR

LouisCastricato: Have it automatically determine what W&B account is logged in. cc @ayulockin

trlx/utils/__init__.py
```diff
@@ -46,6 +49,23 @@ def safe_mkdir(path: str):
...
+def get_distributed_config(accelerator: Accelerator):
+    """
+    Return accelerator distributed config
+    """
+    ds_plugin = accelerator.state.deepspeed_plugin
+    accelerate_config = accelerator.state
+    return {
+        "mixed_precision": accelerate_config.mixed_precision,
+        "num_gpus": accelerate_config.num_processes,
+        "gradient_accumulation_steps": ds_plugin.gradient_accumulation_steps,
```
reciprocated, line 61: I think if ds is not used then `ds_plugin` is None, giving attribute error here
https://github.com/huggingface/accelerate/blob/e4e5611e5d4270a846caf42cba3388e54b83f074/src/accelerate/state.py#L62

maybe some processing of `repr(accelerator.state)` (which is outputed in `accelerate env`) would be equivalent here

configs/ppo_config.yml
reciprocated: it already does so, this option is optional and was added in #78

reciprocated: otherwise it forces people to `wandb disable` as in https://github.com/CarperAI/trlx/issues/106

trlx/utils/__init__.py
Dahoas: I guess it's true in some cases a user may not be using deepspeed so a check should be performed. However with the accelerator.state object there are some items I don't care about (such as the local rank).

Issue #80, opened 2022-11-05
Add LORA support to TRLX


ethankim00:
### 🚀 The feature, motivation, and pitch

LORA and other parameter-efficient methods can provide a number of advantages when finetuning large language models. These methods typically only update a small fraction of the model parameters during finetuning. For example, LORA only trains low-rank reparameterizations of weight matrices during training reducing parameter cost up to 10000 times. 

Key Advantages:

- Reduced storage cost and ability to switch out different adapters for different tasks
- Increased training efficiency and reduced memory requirements since fewer optimizer states need to be tracked

### Additional context

The [OpenDelta](https://opendelta.readthedocs.io/en/latest/index.html) library provides support for LORA and other "delta methods"

```python
model =  AutoModelForCausalLM.from_pretrained(model_base)
from opendelta import LoraModel
delta_model = LoraModel(backbone_model=model, modified_modules=['fc2'])
delta_model.freeze_module(exclude=["deltas", "layernorm_embedding"], set_state_dict=True)
# save only  trained parameters
delta_model.save_finetuned(save_path)
```

LouisCastricato:
cc @Dahoas

Dahoas:
Closing this.

PR #110, opened 2022-11-23
initial commit for trlx LORA support

ethankim00:
Basic support for low rank adaptation.


LouisCastricato:
This should take a similar form to how hydra models are built. It shouldn't be required and directly integrated into ilql or PPO model

LouisCastricato:
https://github.com/CarperAI/trlx/issues/80 Relevant issue.

Dahoas:
@ethankim00 just a gentle push on when you expect to finish this?

cat-state:
cc @Sayanc93

ethankim00:
I can get to it tomorrow or Monday. I'm wondering what the API should be to avoid modifying the model definitions?

cat-state:
> I can get to it tomorrow or Monday. I'm wondering what the API should be to avoid modifying the model definitions?

I think it would be like, instead of modifying the `CausalLMOutputWithCrossAttentions` class or `GPTHydraHeadWithValueModel`, the delta versions could be subclasses, and then the config can treat them as just another architecture to be trained

LouisCastricato:
Circling back around on this.

LouisCastricato:
@cat-state does it make sense to do lora + hydra or just have lora be entirely separate...

ethankim00:
We could have a function to modify the base model of each different model type rather than creating subclasses.

LouisCastricato:
Hm... I differ to better software engineers haha @jon-tow your input would be great here too

jon-tow:
@ethankim00 This looks great! I've made a few changes based on some testing on our cluster. Here's the summary:

* Updates `"gpt_neo"` model type name in the modifier map.

* Fixes layer regex pattern, as the previous one could not capture on ranges with multiple digits, e.g. adapting LORA to a model's 8 through 11 block layers failed since `[8-11]` is an invalid regex character range.

* Moves the delta model modifications to the base trainer to avoid unnecessary duplication.

* Change `_opendelta_available` to `HAS_OPENDELTA` for consistency with other modules (see `HAS_BNB`).

Overall things look very promising. Check out these runs from the PPO sentiments task [here](https://wandb.ai/jon-tow/trlx/reports/trlx-LORA-support--VmlldzozMjgxMzQ0?accessToken=lns57dlkksizbr5n8eyhb5weyr25m7zrfels3hkn24z3pl0p5m1lz0culh5b1yy3). I'm going to begin testing on ILQL and once that's cleared up, we can get this ready for review and a merge.

LouisCastricato:
This looks fantastic! Definitely worth including for the 0.4 release next week. Let's get this merged :)

trlx/data/configs.py
```diff
@@ -37,6 +37,8 @@ class ModelConfig:
     model_path: str
     tokenizer_path: str
     num_layers_unfrozen: int = -1
+    delta_method: str = None
```
LouisCastricato, line 40: We should include lots of documentation about what options can go here.

trlx/trainer/accelerate_base_trainer.py
```diff
@@ -86,8 +81,27 @@ def __init__(self, config, train_mode=True):
                 init_kwargs=init_trackers_kwargs,
             )
 
-        self.opt = self.setup_optimizer()
-        self.scheduler = self.setup_scheduler()
+    def setup_model(self):
```
LouisCastricato, line 84: We didnt have a setup function before?

trlx/utils/modeling.py
```diff
@@ -8,6 +8,19 @@
 import torch.nn.functional as F
 import transformers
 
+try:
```
LouisCastricato, line 11: Its an optional dependency right?

LouisCastricato: Maybe we need a file where we just import all the optional dependencies. I worry this is going to get very gross very fast otherwise.

trlx/utils/modeling.py
```diff
@@ -241,3 +254,207 @@ def update(self, xs: torch.Tensor) -> Tuple[float, float]:
...
+            "self_attention.query_key_value",
+            "self_attention.dense",
+            "mlp.dense_h_to_4h",
+            "mlp.dense_4h_to_h",
+        ],
+    },
+}
+
+
+def generate_layer_regex(config, num_layers_unfrozen: int = -1) -> str:
```
LouisCastricato, line 320: Whats the reason for not wanting to put all of the delta stuff in its own file...

LouisCastricato (approved):
LGTM

trlx/trainer/accelerate_base_trainer.py
jon-tow: Nope, we just created the model and froze parts of it in the constructor. Moving this into a setup function will easily let us add an optional model arg to the `__init__` so users can bring their own models in the future.

trlx/utils/modeling.py
jon-tow: Yeah, good idea! I'll make an issue to track this.

trlx/trainer/accelerate_base_trainer.py
LouisCastricato: Got it.

PR #111, opened 2022-11-23
Ppo z3

Dahoas:
Work in progress integrating zero3 with hydra models for ppo.

Current implementation works for models < 6B but OOMs on 6B.


LouisCastricato:
What are we doing with this PR

PR #114, opened 2022-11-28
Set seed after default config instantiation

jon-tow:
* `set_seed(config.train.seed)` fails if no config is passed to the `train` function as a default config must be created first.


reciprocated (approved):
lgtm

PR #115, opened 2022-11-28
Move queries on the device

reciprocated:
This pr fixes a bug missed by me in 
- #108


jon-tow (approved):
Sorry I didn't catch this in the review! Looks good to me - tested locally 👍

Issue #69, opened 2022-10-25
How to attribute reward to multiple model runs in the same trajectory with PPO


dpaleka:
I want to finetune a base model `M` to maximize a reward `R`, when the model is used inside of a more complex system.
Take a simple example of the setting. The trajectory is as follows: sample `prompt_1` from a dataset of prompts, then
```
prompt1 -> M(prompt1)  = out_1
out_1 -> F(out_1) = prompt_2
prompt_2 -> M(prompt_2) = out_2
out_2 -> R(out_2) = reward
```

where `F : str -> str` and `R : str -> int` are some methods defined in my code.
Is there a way to do this in the current TRLX framework, preferably online with PPO?
Alternative suggestions are welcome.

cat-state:
summarizing from the discord: 
lets say `M = m(params, x)`
then currently you should be able to do online RL with
```
def reward_fn(x):
    return m(params_0, F(x))
```
which would be something like
```
out_i = m(params_i, prompt1)
reward = reward_fn(out_i)
```
This would mean only the first M call is unfrozen, the one after F does not change.

Could you specify the task you are trying to achieve?

dpaleka:
Feedback from a conversation with @LouisCastricato and @Dahoas:
This is not supported in the current TRLX version. The closest thing available is attributing the reward only to the first call of `M`, as in @cat-state's reply.

As of now, TRLX supports only RL setups where all "actions" to attribute the reward to are done before the reward function is called.
The recommended solution is to write your own orchestrator to do this. @LouisCastricato says they might merge this if someone implements it correctly.

The main issue is distributed computing (which is on the roadmap), since the plan is to compute rollouts and rewards on different nodes by default soon. Distributed RL is way easier if the reward function is stateless.


My comments:
If I understand the code correctly, implementing a separate orchestrator to do this right now might not be too difficult. I'm not sure if you should merge it because it might interfere with future parallelization improvements. If I decide to make a PR, I'll try to keep that in mind.

paulbricman:
+1, also interested in something like this.

> As of now, TRLX supports only RL setups where all "actions" to attribute the reward to are done before the reward function is called.

@dpaleka, isn't this already the case in your very first pseudocode snippet? `R` is only called after both `M` calls, not in between, right?

dpaleka:
> 
> > As of now, TRLX supports only RL setups where all "actions" to attribute the reward to are done before the reward function is called.
> 
[...]

PR #117, opened 2022-11-28
experience_fn working branch

dpaleka:
Working PR for experience_fn protocol
From the Discord:

> For the general usecase similar to #69, the following ideas materialized for enhancing ppo_orchestrator.py and trlx.py:
> 
> in addition to reward_fn,  you could have an experience_fn which essentially describes a RL trajectory, and gives the info enough to compute the samples and logprobs that need to be passed to PPORLElement. 
> 
> For example, an experience_fn would describe the model being used to play a textualized version of FrozenLake, or  or any RL setup where the language model is used as a policy.
> 
> The default experience_fn(batch) is just "generate a batch of responses given a batch of prompts using a single call model.generate", which is the current behavior of the PPO orchestator. The API call to trlx.train would remain absolutely the same in this default use case .
> 
> I expect >90% of the user-provided trajectories to have the following properties:
> - "actions" are decided by some series of inferences via model.generate
> - the shape of the trajectory is determined only by the sampled tokens, not by their logprobs
> - the shape of the trajectory does not depend on any intermediate rewards
> - the provided reward_fn takes the generated trajectory in the end, and calculating (possibly token-wise) rewards
> - the user wants to add KL penalties as a part of their reward, and wants this as a built-in default
> 
> There are no other obvious constraints; in particular, the prompts in multiple model inferences do not need to be prefixes of one another, and the user can parse and modify the model outputs in any way they want. This is crucial for e.g. training a language model to play non-text games.
> 
> Given the above, any  experience_fn should work, but could be quite slow depending on whether the model calls can be parallelized.  I expect >50% usecases to have the following property:
> - the model calls in the trajectory can be batched; in particular, any trajectory calls model.generate roughly the same number of times. 
> In those cases, the performance penalty compared to the default experience_fn should scale only by a factor of [number of calls to model_generate in the trajectory].


LouisCastricato:
Any update?

PR #119, opened 2022-11-28
Add ppo randomwalks example

reciprocated:
This pr extends randomwalks example for ppo. Also it changes its reward to be measured in proportion against shortest lengths, which doesn't change much empirically since this is a very easy example, but I find it more interpretable then raw negative lengths.

ILQL: https://wandb.ai/sorry/public/runs/1rl3ls2d
PPO: https://wandb.ai/sorry/public/runs/2vviyvwg

### DIFF
--- a/configs/ilql_randomwalks.yml
+++ b/configs/ilql_randomwalks.yml
@@ -1,23 +1,23 @@
 model:
-  model_path : "randomwalks/1M"
+  model_path: "randomwalks/1M"
   tokenizer_path: "randomwalks/1M"
-  model_type : "AccelerateILQLModel"
+  model_type: "AccelerateILQLModel"
   num_layers_unfrozen: -1
 
 train:
   seq_length: 10
   batch_size: 100
-  epochs: 100
+  epochs: 20
   total_steps: 1000
 
-  lr_init: 1.0e-3
-  lr_target: 1.0e-3
+  lr_init: 2.0e-4
+  lr_target: 2.0e-4
   opt_betas: [0.9, 0.95]
   opt_eps: 1.0e-8
   weight_decay: 1.0e-6
 
   checkpoint_interval: 100000
-  eval_interval: 10
+  eval_interval: 16
 
   pipeline: "PromptPipeline"
   orchestrator: "OfflineOrchestrator"
@@ -25,11 +25,11 @@ train:
 
 method:
   name: "ilqlconfig"
-  tau: 0.7
+  tau: 0.8
   gamma: 0.99
   cql_scale: 0.1
   awac_scale: 1
   alpha: 0.1
-  steps_for_target_q_sync: 1
-  betas: [10]
+  steps_for_target_q_sync: 5
+  betas: [100]
   two_qs: true
--- a/configs/ilql_randomwalks.yml
+++ b/configs/ilql_randomwalks.yml
@@ -1,0 +1,35 @@
+model:
+  model_path : "randomwalks/1M"
+  tokenizer_path: "randomwalks/1M"
+  model_type : "AccelerateILQLModel"
+  num_layers_unfrozen: -1
+
+train:
+  seq_length: 10
+  batch_size: 100
+  epochs: 100
+  total_steps: 1000
+
+  lr_init: 1.0e-3
+  lr_target: 1.0e-3
+  opt_betas: [0.9, 0.95]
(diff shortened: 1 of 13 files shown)


dpaleka:
Would itbe better to host this model [https://huggingface.co/randomwalks](https://huggingface.co/randomwalks) under [https://huggingface.co/CarperAI](https://huggingface.co/CarperAI) or something similar, if the randomwalks repo is the official test example for experimental features as written in [CONTRIBUTING.md](https://github.com/CarperAI/trlx/blob/15f18f8d610f7e79f76d2fb59c2e463fe387202a/CONTRIBUTING.md?plain=1#L52).

reciprocated:
@jon-tow What do you think about Daniel's comment? Also just for the record I changed setup.cfg to let the top import in randomwalks be both visible from the root and when importing from trlx/sweep.py

configs/ilql_randomwalks.yml
```diff
@@ -0,0 +1,35 @@
+model:
```
jon-tow, line 1: Nit: What do you think about moving `ilql_randomwalks.yml` and `ppo_randomwalks.yml` to `examples/randomwalks/configs` (which is something done similarly in the [program synthesis example](https://github.com/CarperAI/trlx/tree/main/examples/experiments/grounded_program_synthesis/config))?

configs/ppo_randomwalks.yml
```diff
@@ -0,0 +1,47 @@
+model:
+  model_path: "randomwalks/1M"
+  tokenizer_path: "randomwalks/1M"
```
jon-tow, line 3: Update the paths to point to CarperAI HuggingFace model path.

configs/ilql_randomwalks.yml
```diff
@@ -0,0 +1,35 @@
...
+  model_path: "randomwalks/1M"
+  tokenizer_path: "randomwalks/1M"
```
jon-tow, line 3: Update the paths to point to CarperAI HuggingFace model path.

examples/randomwalks/randomwalks.py
```diff
@@ -0,0 +1,105 @@
+import networkx as nx
```
jon-tow, line 1: Do we want to add `networkx` as a dependency? I believe this is what you're referring to in [this](https://github.com/CarperAI/trlx/pull/119#issuecomment-1331418733) comment? I don't see it in the `setup.cfg` 😅 - We can also just point users to `pip install networkx` in the `randomwalks/README.md` which might be better. What do you think?

setup.cfg
```diff
@@ -34,6 +34,5 @@ dev =
 
 [options.packages.find]
 exclude =
-    examples*
```
jon-tow, line 37: Good idea!

jon-tow (changes requested):
Nice job! I ran everything locally and it worked great 👍 I left some nits if you could address them before merging.

jon-tow:
> @jon-tow What do you think about Daniel's comment? Also just for the record I changed setup.cfg to let the top import in randomwalks be both visible from the root and when importing from trlx/sweep.py

@reciprocated I think it might be easier to find under CarperAI but ultimately it's up to you given that it's your model 😄 If you want to transfer; feel free!

LouisCastricato:
Host under Carper

reciprocated:
@jon-tow Moved configs and moved the model under carper. Also I added networkx as dependency, since it's pretty [lightweight](https://github.com/networkx/networkx/blob/d82815dba6c8ddce19cd49f700298dc82a58f066/requirements/default.txt#L1) and follows a distinct naming pattern 🙂 

ppo run from fresh git clone: https://wandb.ai/sorry/trlx/runs/2kh7e8q6

configs/ilql_randomwalks.yml
reciprocated: agree

configs/ppo_randomwalks.yml
reciprocated: done

examples/randomwalks/randomwalks.py
reciprocated: I was referring to `from examples.randomwalks import generate_random_walks`, but I also added networkx to make it frictionless

jon-tow (approved):
Looks great to me! 👏

PR #125, opened 2022-12-05
NeMo Integrate

cat-state:
Currently implemented:
* Training for ILQL
* Generation using ILQL (i.e validation)
* Inference script for ILQL 

Future issues:
* More reusable nemo abstraction
* `bf16` appears unstable compared to fp16


trlx/trainer/nn/ilql_models.py
```diff
@@ -33,6 +33,17 @@ def topk_mask(xs: torch.FloatTensor, k: int):
     return torch.where(xs < mintop, -np.inf * torch.ones_like(xs, dtype=xs.dtype), xs)
 
 
+def batched_index_select(x, idxs, dim):
+    """
+    x: [batch, seq, hidden]
+    idxs: [batch, idxs_seq]
+    returns: [batch, idxs_seq, hidden]
```
jon-tow, line 40: Aside: We should probably remove the `torchtyping` dependency since we never use it 😄

trlx/trainer/nn/ilql_models.py
```diff
@@ -33,6 +33,17 @@ def topk_mask(xs: torch.FloatTensor, k: int):
...
+    """
+    # print(f"{mp_rank=} {x.shape=} {idxs.shape=}")
```
jon-tow, line 42: Remove debug print

trlx/trainer/nemo/gpt.py
```diff
@@ -0,0 +1,982 @@
...
+                    stage = "last"
+                else:
+                    stage = "middle"
+
+                pad_by = self.cfg.encoder_seq_length - inputs.shape[1]
+                inputs = torch.nn.functional.pad(
+                    inputs, (0, pad_by), value=self.tokenizer.eos_id
+                )
+
+                # print(f"{pad_by=} {self.cfg.encoder_seq_length=} {inputs.shape=}")
```
jon-tow, line 753: Remove debug print

jon-tow (commented):
Leaving some very minor nits before getting my hands dirty on the cluster with it this weekend. Will report back with another review then.

examples/nemo_ilql_sentiments.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def metric_fn(samples: List[str], **kwargs) -> Dict[str, List[float]]:
+        sentiments = list(map(get_positive_score, sentiment_fn(samples)))
+        return {"sentiments": sentiments}
+
+    imdb = load_dataset("imdb", split="train+test")
+
+    trlx.train(
+        dataset=(imdb["text"], imdb["label"]),
+        eval_prompts=["I don't know much about Hungarian underground"] * 128 * 20,
```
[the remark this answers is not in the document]
cat-state: Oops, this was a hack to make the val dataloader not run out during training. Ill fix it properly

examples/nemo_ilql_sentiments.py
```diff
@@ -0,0 +1,49 @@
...
+def main(hparams={}):
+    config = TRLConfig.update(default_config, hparams)
+
+    sentiment_fn = pipeline(
+        "sentiment-analysis",
+        "lvwerra/distilbert-imdb",
+        top_k=2,
+        truncation=True,
+        batch_size=256,
+        device=-1 if int(os.environ.get("LOCAL_RANK", 0)) == 0 else -1,
```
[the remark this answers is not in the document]
cat-state: I just tried again with 0 but it has to be CPU only for now, otherwise it seems to cause an OOM. I think it should be possible to have the reward function use GPU if its allocated on the right GPUs in the future though

trlx/trainer/nemo/README.md
```diff
@@ -0,0 +1,331 @@
...
+To use a NeMo models in `.nemo` format, un-tar it:
+```
+tar xvf nemo_gpt20B_bf16_tp4.nemo
+```
+This will extract the model weights and the model config.
+
+Then set `train.trainer_kwargs.pretrained_model` to the path to the directory containing the parameters. The model hyperparameters in the `train.trainer_kwargs.megatron_cfg` should match the ones in the model config.
+
+## NeMo Megatron setup
+This setup pertains to the cluster, but should also work if you clone the `r1.13.0` branch of NeMo from https://github.com/NVIDIA/NeMo/tree/r1.13.0 and apex from `nm_v1.13.0` from https://github.com/ericharper/apex.git
```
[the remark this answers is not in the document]
cat-state: Oh interesting, it existed at one time with nemo r1.13 saying to install that. But NeMo main is after r1.14 where they changed that to the other branch. I've successfully been using main for both apex and nemo. Although r1.14 is the supported version.

trlx/trainer/nemo/gpt.py
```diff
@@ -0,0 +1,878 @@
...
+            micro_batch_size=self.cfg.micro_batch_size,
+            global_batch_size=self.cfg.global_batch_size,
+            data_parallel_rank=dp_rank,
+            data_parallel_size=dp_size,
+            drop_last=True,
+        )
+        return torch.utils.data.DataLoader(
+            dataset,
+            batch_sampler=batch_sampler,
+            # For some reason this causes a crash when using >0 workers
```
jon-tow, line 315: If `_reconfigure` for generate is uncommented, as currently is, `num_workers=self.cfg.data.num_workers` seems to work fine. Do you think we should change it back from `0`?

trlx/trainer/nemo/gpt.py
```diff
@@ -0,0 +1,878 @@
...
+        )
+
+        # only the last stages of the pipeline return losses
+        if losses_reduced_per_micro_batch:
+            # average loss across micro batches
+            loss_tensors_list = [
+                loss_reduced["avg"] for loss_reduced in losses_reduced_per_micro_batch
+            ]
+            loss_tensor = torch.concat(loss_tensors_list)
+            loss_mean = loss_tensor.mean()
```
jon-tow, line 487: Can we collect the loss stats here e.g. `loss_q`, `loss_awac`, etc. for logging?

jon-tow:
I think this PR is in a really good state right now. Any updates on the remaining issues you've listed?
- [ ] Verify checkpointing
- [ ] ILQL loss spikes near the start of training (this doesn't seem to be the case now?)

cat-state:
> I think this PR is in a really good state right now. Any updates on the remaining issues you've listed?
> 
>     * [ ]  Verify checkpointing
> 
>     * [ ]  ILQL loss spikes near the start of training (this doesn't seem to be the case now?)

Yeah, I think the last one isn't a case anymore. I'll add a script showing how to load a checkpoint and infer it.

cat-state:
I added the inference script and made the checkpointing save to a reloadable name (doesn't work with metrics with `/` in their name)

jon-tow (commented):
Leaving some *very* minor nits

jon-tow (approved):
Amazing work 🥳 🥳 🥳  Thanks @cat-state !!

PR #133, opened 2022-12-12
Add `bitsandbytes` optimizer support

aicrumb:
Uses bitsandbytes adam optimizer instead of torch, adds very simple gradient accumulation, finetuning only bias/layernorms (tested, works very well and is very fast) and allows for different precisions easier. (sorry for the absolute mass of commits, I was making them on github.com)


trlx/model/accelerate_base_model.py
```diff
@@ -81,8 +88,7 @@ def __init__(self, config, train_mode=True):
                     }
                 },
             )
-
-        self.opt = torch.optim.AdamW(
+        self.opt = bnb.optim.AdamW8bit(
```
jon-tow, line 91: Nice! 

This removes support for vanilla AdamW and some user devices may not be optimized for 8bit (requires Turing or Ampere GPUs). We should introduce an optimizer selection interface in the future but for now, maybe we could add a constructor flag to toggle between 8-bit and vanilla AdamW? What do you think?

trlx/data/configs.py
```diff
@@ -62,6 +66,9 @@ class TrainConfig:
 
     :param batch_size: Batch size for training
     :type batch_size: int
+    
+    :param accumulate_steps: How many forward calls before the optimizer takes a step
+    :type accumulate_steps: int
```
jon-tow, line 71: Re-name this to `gradient_accumulation_steps` to match other APIs ([DeepSpeed](https://www.deepspeed.ai/docs/config-json) and [Accelerate](https://huggingface.co/docs/accelerate/v0.15.0/en/package_reference/cli#accelerate-launch)).

trlx/model/accelerate_base_model.py
```diff
@@ -44,7 +45,13 @@ def __init__(self, config, train_mode=True):
         self.max_length = config.train.seq_length
 
         # Retrieves model equipped for ppo, ilql, etc
-        self.model = self.get_arch(self.config)
+        torch_dtype = {
+            "float32": torch.float32,
+            "float16": torch.float16,
+            "bfloat16": torch.bfloat16
+        }[self.config.model.dtype]
```
jon-tow, line 52: This might be redundant as `accelerate` allows users to specify half floating-point types in the `mixed_precision` config. Have you tried those settings?

jon-tow (changes requested):
Very cool! I've left some comments and change requests 👍

setup.cfg
```diff
@@ -23,6 +23,7 @@ install_requires =
     ray>=2.0.1
     tabulate>=0.9.0
     networkx
+    bitsandbytes
```
jon-tow, line 26: Let's add this as an additional extra require under a tag like `bnb`, so that users could install via `pip install trlx[bnb]`.

jon-tow (changes requested):
CC @cat-state who suggested this change request in Discord - maybe they have more to suggest 🙏

trlx/model/accelerate_base_model.py
```diff
@@ -245,15 +251,17 @@ def learn(self):
...
                     self.accelerator.backward(loss)
                     backward_time = time() - backward_time
 
-                    self.opt.step()
-                    self.opt.zero_grad()
-                    self.scheduler.step()
+                    if self.iter_count % self.config.train.accumulate_steps == 0:
+                        self.opt.step()
+                        self.opt.zero_grad()
+                        self.scheduler.step()
```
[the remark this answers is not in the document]
aicrumb: That'd be great

trlx/model/accelerate_base_model.py
aicrumb: Have not, seems obvious in hindsight, was writing a bit fast

trlx/model/accelerate_base_model.py
aicrumb: That'd probably be best, my only other idea is automatically detecting if bnb is installed but problems arise in bnb compiling without cuda support, and just that giving more control to the user is probably best

jon-tow:
Hey @crumb! Just getting back to this. I think it's best to separate the 3 new features into their own PRs to avoid unwieldy commit histories (if there's an issue in one feature we'd have to revert the other features as well).

Let's make this a PR focused on adding the `bitsandbytes` optimizer support. I believe there might some more thought that needs to go into adding grad accumulation; it's possible `accelerate` handles this in subtly different ways for `deepspeed` integration (which we rely heavily on) and it should be properly tested.

Hope that's alright with you 🙏 

(Also, no worries about the many commits! We squash them before merging)

LouisCastricato:
lets make the bitsandbytes pr

jon-tow:
### Details on latest changes

* We opt to not convert `nn.Embedding` layers into `bnb.nn.StableEmbedding` layers since `trlx` is essentially a fine-tuning library; according to Tim Dettmers:
    > "StableEmbedding layer is only required if the model was pretrained with the StableEmbedding layer."

  See relevant discussion [here](https://github.com/huggingface/transformers/issues/14839#issuecomment-1003431668).

  * Instead, we take the more general approach of forcing `nn.Embedding` layers into 32-bit precision following advice from Tim Dettmers: 
 
     > "for pretrained models, the best would be to use the 32-bit optimized embedding layer (`bnb.nn.Embedding`) and no layer norm if the pretrained model was not trained with a [StableEmbedding] layer norm." 

    Again, relevant discussion [here]( https://github.com/huggingface/transformers/issues/14819#issuecomment-1016017746).

### Reports
* In practice, we can see ~13% memory savings for GPT-J when trained with 8 **un**frozen layers. Note that large memory savings occur only when a large portion of the model is unfrozen (`num_layers_unfrozen` is approx. some large fraction of total model layers).
  * There seems to be a slight penalty for using 8-bit optimizers but nothing significantly divergent - can possibly be overcome with improved hyper-parameter tuning.
  * Relevant `wandb` report: [here](https://wandb.ai/carperai/trlx/reports/trlx-Add-bitsandbytes-optimizer-support-133--VmlldzozMjU5OTQx)

CC @LouisCastricato

LouisCastricato:
Looks to me, let's merge.

LouisCastricato (approved):

jon-tow (approved):

reciprocated:
hey @jon-tow, your wandb report (https://wandb.ai/carperai/trlx/reports/trlx-Add-bitsandbytes-optimizer-support-133--VmlldzozMjU5OTQx) doesn't render for me for some reason (there are no data in charts as if the run set is empty),  is it again some wandb's shenanigans or just me 😴

LouisCastricato:
It was rendering yesterday. I think wandb is just being weird.

jon-tow:
@reciprocated those were `jon-tow` shenanigans 😅 Yesterday I discovered deleting a run also deletes its data from reports (which makes sense in hindsight). I re-ran them from my personal wandb account - try this link: https://wandb.ai/jon-tow/trlx/reports/trlx-Add-bitsandbytes-optimizer-support-133--VmlldzozMjY1MzI1

reciprocated:
@jon-tow that link gives 404 for me 😓 you have to share a "magic" link instead

jon-tow:
@reciprocated This oughta do it: https://wandb.ai/jon-tow/trlx/reports/trlx-Add-bitsandbytes-optimizer-support-133--VmlldzozMjY1MzI1?accessToken=4t40dgib8rzs1qb6kqgsx8yuqhp5csfc2dsmxpz0qdwv2justj4yer44qwwtg7qp

reciprocated:
@jon-tow Thanks 🙏

PR #135, opened 2022-12-14
Add `OptimizerConfig` and `SchedulerConfig`

jon-tow:
This PR adds the following configuration updates:

- Adds an `OptimizerConfig` dedicated to customizing optimizer specifications.
    - Previously users were hard constrained to `torch.optim.AdamW`.
    - This should make it easier to add other optimizers like 8-bit optimizers
      introduced in PR #133

- Adds a `SchedulerConfig` for customizing the learning rate scheduler.
    - Previously users were hard constrained to `torch.optim.lr_scheduler.CosineAnnealingLR`; this opens possibilities for other schedulers like `torch.optim.lr_scheduler.OneCycleLR` if desired from the community 

Notes:

- This PR formats the config YAML files throughout the repository to be more
  consistent.

`wandb` reports: [PPO/ILQL Sentiments](https://wandb.ai/jon-tow/trlx/reports/trlx-Add-OptimizerConfig-and-SchedulerConfig-135--VmlldzozMTQ1NTQw?accessToken=b3quf5f60zz3quxch85w2p5avobeop5hb73sgydoadfsu89get5y0lykzeqjmv4p)

@ reviewer: I found these updates useful to me; if this feature is not desired or designed appropriately feel free to suggest edits or close.

### DIFF
--- a/trlx/utils/__init__.py
+++ b/trlx/utils/__init__.py
@@ -1,6 +1,7 @@
 import os
 import random
 import time
+from enum import Enum
 from functools import reduce
 from typing import Any, Iterable, List, Dict
 from dataclasses import is_dataclass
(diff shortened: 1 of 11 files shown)


trlx/utils/__init__.py
```diff
@@ -86,6 +87,50 @@ def get_distributed_config(accelerator: Accelerator):
     return dist_config
 
 
+class OptimizerNames(Enum):
```
LouisCastricato, line 90: Why are we not using registers...?

LouisCastricato (approved):

trlx/utils/__init__.py
jon-tow: Adding a registry for `str` types seemed like a pre-mature abstraction given that the source is local to a single file. What exactly are you suggesting?

LouisCastricato: Fair, Dw then. Ready for merging imho.

PR #136, opened 2022-12-14
add: load function for AccelerateRLModel

dongs0104:
# save and load are being done in different ways.

After performing the example, the model that was save at `ckpts`
it call https://github.com/CarperAI/trlx/blob/1a3461d592a567409e11e3c47b5b355bf219449f/trlx/model/accelerate_base_model.py#L139
when I use `ckpts` weight it call https://github.com/CarperAI/trlx/blob/1a3461d592a567409e11e3c47b5b355bf219449f/trlx/model/__init__.py#L119

### DIFF
--- a/trlx/model/accelerate_base_model.py
+++ b/trlx/model/accelerate_base_model.py
@@ -128,20 +128,12 @@ class AccelerateRLModel(BaseRLModel):
                 input_ids=input_ids, attention_mask=attention_mask, **kwargs
             )
 
-    def get_components(self) -> Dict[str, Any]:
-        components = (
-            {"model": self.model, "opt": self.opt, "scheduler": self.scheduler}
-            if self.train_mode
-            else {"model": self.model}
-        )
-        return components
-
     def save(self, directory=None):
         """Creates checkpoint of optimizer, scheduler and a model"""
         self.accelerator.save_state(directory or self.config.train.checkpoint_dir)
 
     def load(self, directory=None):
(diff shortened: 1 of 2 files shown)


trlx/model/accelerate_base_model.py
```diff
@@ -140,6 +140,10 @@ def save(self, directory=None):
         """Creates checkpoint of optimizer, scheduler and a model"""
         self.accelerator.save_state(directory or self.config.train.checkpoint_dir)
 
+    def load(self, directory=None):
+        """load checkpoint of optimizer, scheduler and a model"""
```
jon-tow, line 144: Nit: Capitalize the first `l` in the `"""load ...` docstring

trlx/model/accelerate_base_model.py
```diff
@@ -140,6 +140,10 @@ def save(self, directory=None):
...
+        self.accelerator.load_state(directory or self.config.train.checkpoint_dir)
```
jon-tow, line 145: Delete the `get_components` override here: https://github.com/CarperAI/trlx/blob/ee6e58c5417ca56e0cb93a1eb701b61889f0c6e4/trlx/model/accelerate_base_model.py#L131-L138
since it will never get called after your `load` update.

jon-tow (changes requested):
Good catch and thanks for the fix! I've left a nit and change request.

LouisCastricato (approved):

jon-tow (approved):
Thanks again @dongs0104 !

PR #140, opened 2022-12-17
Freeze hydra model branches 

jon-tow:
This PR updates the `ModelBranch` classes used in the PPO hydra architectures to freeze all parameters by default.
Currently, only the transformer blocks are frozen. Note that the `ModelBranch.lm_head`s are never frozen but the
`lm_head` they are `deepcopy`ed from are instead.

wandb report: [PPO](https://wandb.ai/jon-tow/trlx/reports/trlx-fix-frozen-branch-PR--VmlldzozMTY4Njc1)


Dahoas (approved):
Looks good! Sorry for the delay.

PR #141, opened 2022-12-17
Fix seeding per process

reciprocated:
This pr will grant each process to have a seed of its own
Tested with `accelerate==0.15.0`
[PPO difference](https://wandb.ai/sorry/public/reports/Fix-seeding-per-process-141---VmlldzozMTcwMjgz)


jon-tow (approved):
Looks good to me 👍

PR #143, opened 2022-12-17
Set device-specific seeding with global rank

jon-tow:
I overlooked this in the review of #141. With `LOCAL_RANK`, each device in a single node gets a unique seed but we probably want unique seeds across nodes as well for true device-specific seeding.


reciprocated (approved):
it is correct 🙏

PR #145, opened 2022-12-21
Add T5 model 

PhungVanDuy:
- [x] Support T5 model to trlx
- [x] Fix tokenizer truncated problem, fix PPO state values, rewards at rollouts typo
- [x] Add some utils function: save the best model, save the model to hf, and separate arguments for rollouts and evaluation. 
- [x] Examples for T5: 
 CNN/Dailly mail: https://wandb.ai/pvduy/trlx/runs/lx4iq23e
PPO fixed  sentiments (GPT2): https://wandb.ai/pvduy/trlx/runs/2cyo46k4


Dahoas (commented):
- I see you're adding a bunch of new tasks in your pr (which is great!) but they should probably be separated out into other prs if possible 
- Do you have a wandb you can share?
- I would suggest not freezing anything first (on a very small model with a single gpu) to make sure the algo is right

examples/ds_config_trlx_neoj.json
```diff
@@ -0,0 +1,22 @@
+{
```
Dahoas, line 1: You'll probably want to put this under configs when finished

examples/ds_config_trlx_t5.json
```diff
@@ -0,0 +1,22 @@
+{
```
Dahoas, line 1: Same here (move under configs)

examples/reward_model.py
```diff
@@ -0,0 +1,108 @@
+from transformers import GPT2LMHeadModel, GPT2Tokenizer, GPT2Model, PreTrainedModel, AutoModelForCausalLM, GPT2PreTrainedModel, GPT2Model
```
Dahoas, line 1: We will probably want to keep the reward model code somewhere separate. Likely in the magiCarp repo.

examples/summarize_dataset.py
```diff
@@ -0,0 +1,110 @@
+import torch
```
Dahoas, line 1: Eventually we'll want to put this dataset onto huggingface

PhungVanDuy:
@Dahoas  @LouisCastricato , this is example T5 worked for CNN-Dailymail dataset: https://wandb.ai/pvduy/trlx/runs/8q3skf8p

trlx/pipeline/offline_pipeline.py
```diff
@@ -17,16 +17,25 @@ class PromptPipeline(BasePipeline):
 
     def __init__(self, prompts: List[str], max_prompt_length: int, tokenizer=None):
         super().__init__()
-
-        if tokenizer:
+        if tokenizer and 't5' in tokenizer.name_or_path:
+            model_inputs = tokenizer(prompts, truncation=True, padding="max_length", max_length=max_prompt_length)
```
jon-tow, line 21: Is there any reason to truncate earlier tokens:
`<EARLY tokens> ... <RECENT tokens> ➡️ {EARLY tokens}`
instead of:
`<EARLY tokens> ... <RECENT tokens> ➡️ {RECENT tokens}`
Aren't we also `left` padding which suggests we can end up with more pad tokens than necessary this way?

I believe you mentioned this issue in the Discord but I'm not sure why `seq2seq` architectures would require a different truncation here 😅 Could you explain for reference? Thanks!

examples/ppo_config_cnn_daily.yml
```diff
@@ -0,0 +1,57 @@
...
+
+  pipeline: "PromptPipeline"
+  orchestrator: "PPOOrchestrator"
+  trainer: "AcceleratePPOTrainer"
+
+model:
+  model_path:  "google/flan-t5-large" #  "lvwerra/gpt2-imdb" #"lvwerra/gpt2-imdb"
+  model_arch_type: "encoder_decoder"
+  tokenizer_path:  "google/flan-t5-large" #  "lvwerra/gpt2-imdb" #"google/flan-t5-large" # "lvwerra/gpt2-imdb"
+  num_layers_unfrozen: 2
```
jon-tow, line 18: Remove the comments in the `*_path`s

trlx/orchestrator/ppo_orchestrator.py
```diff
@@ -103,54 +114,98 @@ def make_experience(self, num_rollouts: int = 1024, iter_count: int = 0):
...
-
+            if self.trainer.config.model.model_arch_type == "encoder_decoder":
+                for ix in range(n):
+                    rs = rewards[ix]
+                    rs[-1] = scores[ix]
+                    all_rewards[ix] = rs
+            else:
+                for ix in range(n):
+                    rs = rewards[ix][start : ends[ix]]
+                    if len(rs) == 0: # fix for empty responses
```
jon-tow, line 205: Ah, nice catch. CC @reciprocated : this was the issue I came across a while back on cases where `start == ends[idx]`; were just empty responses.

jon-tow (changes requested):
Looks awesome! I've left some feedback to be addressed. Very excited about this 👍

PhungVanDuy:
> Looks awesome! I've left some feedback to be addressed. Very excited about this 👍

Thank you for your great comment, I will follow up and fix that.

trlx/data/configs.py
```diff
@@ -36,6 +36,7 @@ class ModelConfig:
 
     model_path: str
     tokenizer_path: str
+    model_arch_type: str = "decoder"
```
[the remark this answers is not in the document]
PhungVanDuy: yeah, I do agree with  "causal"/"seq2seq", I will change it in the next commit.

trlx/utils/modeling.py
```diff
@@ -12,9 +12,9 @@
 def make_head(n_embd: int, out: int) -> nn.Sequential:
     """Returns a generic sequential MLP head."""
     return nn.Sequential(
-        nn.Linear(n_embd, n_embd * 2),
+        nn.Linear(n_embd, n_embd * 2, bias=False),
         nn.ReLU(),
-        nn.Linear(n_embd * 2, out),
+        nn.Linear(n_embd * 2, out, bias=False),
```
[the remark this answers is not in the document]
PhungVanDuy: This setup in a few experiments will keep bias in the newest code.

trlx/orchestrator/ppo_orchestrator.py
```diff
@@ -68,16 +68,27 @@ def make_experience(self, num_rollouts: int = 1024, iter_count: int = 0):
...
+                        
             texts = self.trainer.tokenizer.batch_decode(
                 samples, skip_special_tokens=True
             )
+
+            if self.trainer.config.model.model_arch_type == "encoder_decoder":
+                articles = self.trainer.tokenizer.batch_decode(
+                    batch.input_ids, skip_special_tokens=True
+                )
+                texts = [f"{article}<sep>{response}" for article, response in zip(articles, texts)]
```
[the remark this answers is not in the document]
PhungVanDuy: Great idea, will do it.

examples/ppo_config_cnn_daily.yml
PhungVanDuy: Will remove it in next updates. Thank you.

trlx/pipeline/offline_pipeline.py
PhungVanDuy: Here are several points:
1. In the summary setup for the TL;DR dataset, the prompt input format of the dataset to train SFT is like this: 
```SUBREDDIT: r/relationships TITLE: Me [25F] with my friend [27F], she hasn't made .....```
After that when I do PPO by `trlx`, the prompt changed because in long text cases, keywords like: "SUBREDDIT", and "TITLE" ... could be removed, and with my experiments, it affected the result.
2. When you use `flant5` sometime you will add the instruction at the beginning like:
- Summarize this document: (CNN) The football match..... ---> same as the case above "Summarize this document" instruction could be removed.  
3. By default some tokenizers default truncated by `right` but in our implementation, we do truncate by `left` this may not be clear to users. As I know `RL4LMs` also have options in the configure file to choose `padding_side` and `truncation_side`, users can choose to fit with their setup.

Comment (reciprocated) [reply trlx/pipeline/offline_pipeline.py:21]: You're right Duy, it's not specific to `seq2seq` but rather to the task and I was about to suggest moving it as a config option as in `RL4LM`

Comment (PhungVanDuy) [reply trlx/pipeline/offline_pipeline.py:21]: Thank you. I will try to add it to the config for this PR. :D

**Comment by PhungVanDuy** [comment]
Fixed PPO for T5 (https://wandb.ai/pvduy/trlx/runs/1n31fb6a). The fix for GPT-J still running on the OpenAI summarization dataset to check. Please review this @reciprocated  @LouisCastricato

File: examples/ppo_sentiments.py [hunk 11]
```diff
@@ -43,13 +43,15 @@ def reward_fn(samples: List[str]) -> List[float]:
...
+    imdb = load_dataset("imdb", split="train")
     prompts = [" ".join(review.split()[:4]) for review in imdb["text"]]
+    imdb = load_dataset("imdb", split="test")
+    val_prompts = [" ".join(review.split()[:4]) for review in imdb["text"]]
 
     return trlx.train(
         reward_fn=reward_fn,
         prompts=prompts,
-        eval_prompts=["I don't know much about Hungarian underground"] * 64,
+        eval_prompts=val_prompts[0:1000],
```
LouisCastricato, line 54: Lmao it was just a matter of time

examples/trlx_t5_summ_daily_cnn.py
```diff
@@ -0,0 +1,68 @@
...
+from datasets import load_dataset
+from tqdm import tqdm
+from transformers import AutoTokenizer
+
+import trlx
+from trlx.data.configs import TRLConfig
+
+meteor = evaluate.load("meteor")
+
+if __name__ == "__main__":
```
LouisCastricato, line 13: This this is an example we probably should have lots of comments.

trlx/orchestrator/ppo_orchestrator.py
```diff
@@ -40,6 +40,7 @@ def __init__(
 
         if not hasattr(self.trainer.model, "frozen_head"):
             self.ref_model = self.trainer.get_arch(self.trainer.config)
+            self.ref_model.to(self.trainer.accelerator.device)
```
LouisCastricato, line 43: @Dahoas

examples/t5_sentiment_train_lm.py
```diff
@@ -0,0 +1,208 @@
...
+
+
+if __name__=="__main__":
+    tokenizer = AutoTokenizer.from_pretrained("google/flan-t5-large")
+    model = AutoModelForSeq2SeqLM.from_pretrained("flan-t5-review")
+    train_dataset = MethodDataset(tokenizer,'train')
+    valid_dataset = MethodDataset(tokenizer,'val')
+    
+    
+    class Hparam():
```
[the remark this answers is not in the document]
PhungVanDuy: Seem you are reviewing old commit. This file removed from this PR.

examples/ppo_sentiments.py
PhungVanDuy: I will remove sentiment examples from this PR

examples/summarize_dataset.py
jon-tow: (Possibly) Relevant: https://huggingface.co/datasets/openai/summarize_from_feedback

PhungVanDuy: Actually, I used this dataset before they were public to hf, but that was for the RLHF blog post not for this PR.

trlx/utils/modeling.py
```diff
@@ -275,3 +275,87 @@ def update(self, xs: torch.Tensor) -> Tuple[float, float]:
...
+    zipped = zip(output_ids, output_scores)
+    # sort by score in ascending order
+    zipped = sorted(zipped, key=lambda x: x[1], reverse=True)
+
+    # unzip
+    output_ids, output_scores = zip(*zipped)
+
+    # and return the best one
+    outputs = []
+    top_n = 1
```
LouisCastricato, line 356: needs to be a kwarg

jon-tow:
@PhungVanDuy If the best-of-n sampling is not required for T5 support, I think it's best to create a separate PR for it to allow for proper testing (I found some basic issues in the previous commits). Then we could better review that without further complicating this PR.

PhungVanDuy:
> @PhungVanDuy If the best-of-n sampling is not required for T5 support, I think it's best to create a separate PR for it to allow for proper testing (I found some basic issues in the previous commits). Then we could better review that without further complicating this PR.

You are right @jon-tow , I am removing from this PR we can consider merging this PR today, some bugs with the current main branch should be fixed by this PR. cc @LouisCastricato

trlx/orchestrator/ppo_orchestrator.py
Dahoas: Have we verified this works? I recall accelerate freezing up if I started putting multiple models on gpu(though this could've just been the sentiment pipeline we were using for sentiments task)

PhungVanDuy: @Dahoas  yes that works. 
e.g. https://wandb.ai/pvduy/trlx/runs/2wgpt4im

PhungVanDuy: I do agree with the larger model we should distribute multiple models on multiple gpus, but for this, I think we should keep it on GPUs rather than CPUs, they are super slow.

trlx/utils/modeling.py
```diff
@@ -96,6 +113,18 @@ def hf_get_causal_base_model(model: transformers.AutoModelForCausalLM) -> nn.Mod
...
+    """Returns the encoder-decoder backbone of the specified HuggingFace transformers
+    model.
+    NOTE: Different model configurations have different encoder-decoder attribute
+    names.
+        - model.encoder: (BartConfig)
+        - model: (MBartConfig)
+    """
+    decoder_attrs = ("shared", "encoder", "decoder")
+    return findattr(model, decoder_attrs)
+
```
jon-tow, line 126: This is unused. Is this intentional? If so, let's remove it.

trlx/utils/modeling.py
```diff
@@ -126,6 +155,11 @@ def hf_get_causal_hidden_layers(model: nn.Module) -> Tuple[nn.Module]:
     return findattr(model, hidden_layers_attrs)
 
 
+def hf_decoder_hidden_layers(model: nn.Module) -> Tuple[nn.Module]:
+    hidden_layers_attrs = "decoder.block"
+    return findattr(model, hidden_layers_attrs)
```
jon-tow, line 160: This is unused. Is this intentional? If so, let's remove it.

trlx/trainer/nn/ppo_models.py
```diff
@@ -404,6 +405,214 @@ def forward(
...
+            decoder_head_mask=decoder_head_mask,
+            cross_attn_head_mask=cross_attn_head_mask,
+            use_cache=use_cache,
+            output_attentions=output_attentions,
+            output_hidden_states=output_hidden_states,
+            return_dict=return_dict,
+        )
+        t5_outputs = self.base_model(**forward_kwargs)
+        lm_logits = t5_outputs.logits
+        last_hidden_state = t5_outputs.decoder_hidden_states[-1]
```
jon-tow, line 514: Nice use of the `t5_outputs` data struct 👍

jon-tow (changes requested):
Leaving some tiny final comments and change requests 🙏

trlx/utils/modeling.py
PhungVanDuy: I have a plan for when we have options to freeze the encoder, and keep the decoder but it is ok to remove it now.

trlx/trainer/nn/ppo_models.py
```diff
@@ -124,6 +124,7 @@ class PPOConfig(MethodConfig):
     ref_std: Optional[float]
     cliprange_reward: float
     gen_kwargs: dict
+    gen_inference_kwargs: Optional[dict] = None
```
jon-tow, line 127: Double-checking: In which cases will `gen_kwargs` and `gen_inference_kwargs` be different? (Just don't want to duplicate fields containing the same info)

PhungVanDuy: In some cases, we need sampling for the exploration phase and beam search or something else for the inference that is why I want to add this one like RM4LMs config. Is it reasonable?

trlx/utils/modeling.py
jon-tow: Oooo that sounds interesting. Thanks for bringing it up!

trlx/trainer/nn/ppo_models.py
PhungVanDuy: e.g. https://github.com/allenai/RL4LMs/blob/main/scripts/training/task_configs/narrative_qa/t5_ppo_on_supervised.yml

PhungVanDuy: by default, if people do not settings this argument it will be the same as `gen_kargs`.

jon-tow: Ah okay. So one setting is for evaluation/inference and the other for the actual in-training policy. Thanks! We should also add a docstring to the `PPOConfig` below the `gen_kwargs` 😅  https://github.com/CarperAI/trlx/blob/220c8f3a3c91738f26e8522effb90b5caa119df4/trlx/trainer/nn/ppo_models.py#L107-L108

reciprocated: Maybe it would be a little clearer to distinct between the two with some renaming. For example if `gen_kwargs` are kept for evaluation, but new kwargs can be called `gen_experience_kwargs` if beam search/different options are needed specifically for PPO exploration

trlx/orchestrator/ppo_orchestrator.py
```diff
@@ -105,52 +121,117 @@ def make_experience(self, num_rollouts: int = 1024, iter_count: int = 0):  # noq
...
             # Compute rewards
-            rewards = -self.trainer.kl_ctl.value * (logprobs - ref_logprobs)
             all_rewards = [None] * n
+
             for ix in range(n):
-                rs = rewards[ix][start : ends[ix]]
-                rs[-1] = scores[ix]
+                rs = rewards[ix]
+                if len(rs) == 0:
+                    rs = torch.tensor([0.0])
```
reciprocated, line 233: Should we penalize empty responses? Also I can't understand how it's possible to have those unless `max_new_tokens == 0` 🤔

trlx/trainer/nn/ppo_models.py
PhungVanDuy: That's a cool idea. I can change it but just make sure we have some notes for users they also need to change their configure then.

trlx/orchestrator/ppo_orchestrator.py
PhungVanDuy: Yes, this is a complete exception case, but I got it a few times when I ran PPO sentiments.

reciprocated: Yeah Jon has said that he also experienced it. I wonder if the case of it is unknown it may be a symptom of some other bug elsewhere

jon-tow: @reciprocated the only case I can think of that'd lead to an empty response is when `len(query)` is larger than the `generate` method's `min_length` arg, which defaults to `10`, and the model so happens to output the `eos_token` on its first sample. (Note that with causal models the `min_length` constraint includes the length of the context (query) meaning it won't actually have an effect on the generations if the min condition is already met by the context size).

LouisCastricato: I do not think it is a serious issue. We can just throw an error if this min_length thing comes up. I've never seen this in practice when I set min length correctly. (Perhaps we should add an extra parameter called min_new_length...? We should upstream to HF transformers though)

jon-tow (approved):

jon-tow (commented):
Leaving some notes for myself to address in the future.

PR #147, opened 2022-12-22
Logging learning rate

leshanbog:
In this PR, learning rate is added to stats. Now it is easy to verify the LR scheduler work in wandb:

image

### DIFF
--- a/trlx/trainer/accelerate_base_trainer.py
+++ b/trlx/trainer/accelerate_base_trainer.py
@@ -282,7 +282,7 @@ class AccelerateRLTrainer(BaseRLTrainer):
                     stats["time/forward"] = forward_time
                     stats["time/backward"] = backward_time
                     for group_number, lr in enumerate(self.scheduler.get_last_lr()):
-                        stats[f"learning_rate/group_{group_number}"] = lr
+                        stats[f"learning_rate_group_{group_number}"] = lr
 
                     if self.iter_count % self.config.train.eval_interval == 0:
                         results = self.evaluate()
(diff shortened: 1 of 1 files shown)


LouisCastricato:
What does group number refer to?

trlx/trainer/accelerate_base_trainer.py
```diff
@@ -281,6 +281,8 @@ def learn(self):
 
                     stats["time/forward"] = forward_time
                     stats["time/backward"] = backward_time
+                    for group_number, lr in enumerate(self.scheduler.get_last_lr()):
+                        stats[f"learning_rate/group_{group_number}"] = lr
```
jon-tow, line 285: Currently, we're logging a ton of metrics: 
image
so to avoid adding extra clutter, I think it might be best to keep the LR dynamics tracked in `Charts` by using the following stats key instead: `stats[f"learning_rate_group_{group_number}"] = lr`

jon-tow (changes requested):
Very cool!

trlx/trainer/accelerate_base_trainer.py
leshanbog: Agree. Moved it to charts

image

leshanbog:
> What does group number refer to?

An index number of a particular [parameter group](https://pytorch.org/docs/stable/optim.html#per-parameter-options). In most cases there will be only one but we will also plot correctly in case of several groups

jon-tow (approved):
Thanks for the new feature! LGTM

PR #150, opened 2022-12-24
Linear LR scheduler

leshanbog:
In this PR:

- Linear LR scheduler is added. It was used in InstructGPT. From the [paper](https://arxiv.org/abs/2203.02155):

image

-  Tested with config:

```yaml
scheduler:
  name: "linear"
  kwargs:
    start_factor: 0.1
    end_factor: 1.0
    total_iters: 64
```

image

- Some unused code is removed

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1,8 +1,8 @@
+import accelerate
 import pytest
 import torch
 import transformers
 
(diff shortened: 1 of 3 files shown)


jon-tow (approved):
Thanks for another great PR :)

PR #151, opened 2022-12-26
Upgrade code quality

cclauss:
Upgrade Linux, GitHub Actions, and Python to current versions.

### DIFF
--- a/.github/workflows/code_quality.yml
+++ b/.github/workflows/code_quality.yml
@@ -4,10 +4,10 @@ on: [pull_request]
 
 jobs:
   code-quality:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-python@v2
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
         with:
-          python-version: 3.9
+          python-version: 3.x
       - uses: pre-commit/action@v2.0.3
--- a/.github/workflows/code_quality.yml
+++ b/.github/workflows/code_quality.yml
@@ -4,10 +4,15 @@ on: [pull_request]
 
 jobs:
   code-quality:
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.9", "3.11"]
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: actions/setup-python@v4
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v4
         with:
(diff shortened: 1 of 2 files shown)


LouisCastricato:
Hey can you please add a description to the PR. A few sentences about what it is that you're suggesting.

LouisCastricato (approved):

.pre-commit-config.yaml
```diff
@@ -2,7 +2,7 @@
 # See https://pre-commit.com/hooks.html for more hooks
 # This should be the _latest_ version of python supported by us
 default_language_version:
-  python: python3.9
+  python: python3.11
```
jon-tow, line 5: Revert this to `python3.9`. We've standardized this version internally across core contributors and do not find any benefit in upgrading now.

.github/workflows/code_quality.yml
```diff
@@ -4,10 +4,10 @@ on: [pull_request]
...
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-python@v2
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
         with:
-          python-version: 3.9
+          python-version: 3.x
```
jon-tow, line 12: Replace the python-versioning with a matrix config to ensure code quality across the earliest supported version (`3.9`) and the latest python version (`3.11`). E.g.
```yml
    strategy:
      matrix:
        python-version: ["3.9", "3.11"]

    steps:
      - uses: actions/checkout@v3
      - name: Set up Python ${{ matrix.python-version }}
        uses: actions/setup-python@v4
        with:
          python-version: ${{ matrix.python-version }}
```

jon-tow (changes requested):
Thanks, @cclauss. I've left some change requests to address. The upgrade isn't required at this moment but we could merge under the conditions I've outlined.

.pre-commit-config.yaml
cclauss: What’s new in Python 3.11…
> Python 3.11 is between 10-60% faster than Python 3.10. On average, we measured a 1.25x speedup on the standard benchmark suite.

.github/workflows/code_quality.yml
cclauss: The deprecated pre-commit action cannot handle multiple Python versions when the Python version is set in the pre-commit config.
> RuntimeError: failed to find interpreter for Builtin discover of python_spec='python3.9'

jon-tow: Ah, that's annoying. I wonder if updating the pre-commit rev would fix this?

PR #39, opened 2022-10-18
Add isort flake8

cat-state:
TODO: make repo flake8 compatible

PR #152, opened 2022-12-27
Update `pre-commit` version and add `isort`

jon-tow:
This PR adds the following `pre-commit` updates:

* ~~Update `pre-commit-hook` to a more recent version.~~
* Adds `black` formatting to `tests` directory as it was never updated for the name change from `unittests`➡️`tests`
* Adds `isort` to the pre-commit action runner. This is to avoid waiting on #39 for the more involved `flake8` checks which will require working around `torchtyping`.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,11 +1,8 @@
 # See https://pre-commit.com for more information
 # See https://pre-commit.com/hooks.html for more hooks
-# This should be the _latest_ version of python supported by us
-default_language_version:
-  python: python3.9
 repos:
 -   repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v3.2.0
+    rev: v4.4.0
     hooks:
     -   id: trailing-whitespace
     -   id: end-of-file-fixer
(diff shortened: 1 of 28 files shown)


trlx/utils/loading.py
```diff
@@ -14,6 +9,11 @@
 from trlx.pipeline import _DATAPIPELINE
 from trlx.pipeline.offline_pipeline import PromptPipeline
 
+# Register load models via module import
```
LouisCastricato, line 12: Somtimes isort fucks up files like this. Maybe loading should have isort skip?

.pre-commit-config.yaml
```diff
@@ -1,11 +1,8 @@
 # See https://pre-commit.com for more information
 # See https://pre-commit.com/hooks.html for more hooks
-# This should be the _latest_ version of python supported by us
-default_language_version:
-  python: python3.9
 repos:
 -   repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v3.2.0
+    rev: v4.4.0
```
LouisCastricato, line 5: Why the change?

jon-tow: Ah, no need actually. The hooks we use haven't been updated between `3.2.0` and `4.4.0`. Reverting.

trlx/utils/loading.py
jon-tow: I actually think it's doing the right thing here since, previously these were `.models` imports and never were re-sorted as `.trainers`. But I can skip this file if you think it's best. Let me know!

LouisCastricato (approved):

cat-state (approved):

PR #156, opened 2023-01-02
Update ILQL details

reciprocated:
This pr
- adds tensor statistics to ILQL training
- changes offline training api by removing `split_token` and instead delegating preprocessing per example
- confines optional crossentropy loss to logits from completions

https://api.wandb.ai/report/sorry/t4w326pt
https://api.wandb.ai/report/sorry/93xza5j0

The difference in randomwalks is due to removing `<bos>` from samples and specifying starting nodes directly instead


trlx/orchestrator/offline_orchestrator.py
```diff
@@ -1,76 +1,125 @@
+from typing import List, Union
+
+import numpy as np
 import torch
 
 from trlx.orchestrator import Orchestrator, register_orchestrator
 from trlx.pipeline.offline_pipeline import ILQLRolloutStorage
 
 
+def tokenize_dialogue(
```
LouisCastricato, line 10: Why are we adding dialogue related functionality to the core api?

LouisCastricato:
Where exactly does this fix the memory issue?

trlx/orchestrator/offline_orchestrator.py
```diff
@@ -1,76 +1,125 @@
...
+        for phrase in reversed(dialogue):
+            tokens = tokenizer(phrase).input_ids[-ctx_length:]
+            ctx_length -= len(tokens)
+            out.insert(0, tokens)
+            if ctx_length == 0:
+                break
+
+        # in case of odd number of phrases (possibly due to truncation)
+        # since the first phrase always has to be a prompt, force it to be <bos>
+        if len(out) & 1:
```
cat-state, line 32: kids these days don't know bitwise ops.. maybe `len(out) % 2 == 1` would be clearer for most people

trlx/orchestrator/offline_orchestrator.py
cat-state: I think having this func in trlx makes sense, it makes the completion split based preproc for ilql a bit clearer although we should probably pair it with a multi-turn example to show it working

examples/ilql_sentiments.py
```diff
@@ -36,7 +36,6 @@ def metric_fn(samples: List[str]) -> Dict[str, List[float]]:
     imdb = load_dataset("imdb", split="train+test")
 
     trlx.train(
-        "gpt2",
```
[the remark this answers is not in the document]
jon-tow: I like this change. Previously, if your ILQL config set the `model_path` option it would be overridden by this `"gpt2"` arg. It's led to unexpected behavior on my end whereby editing the example ILQL config's model path doesn't actually update the model.

reciprocated:
> Where exactly does this fix the memory issue?

Implicitly it fixes the issue by removing `split_token` which was easy to abuse and turn the whole sample into an output over which q/v losses have to be computed. Also cross-entropy is now computed only on a few output logits (as per Charlie's comment) instead of the whole sample. From results of both I can now train on @AlekseyKorshuk's data referenced in the issue with the same `gpt-xl` no layers frozen, mixed precision, batch size 8, context length 1024, on a single A100 without OOMs.

https://wandb.ai/sorry/public/runs/17v1c8st

```python
import yaml
import trlx
from trlx.data.configs import TRLConfig
from datasets import load_dataset

default_config = yaml.safe_load(open("configs/ilql_dalio.yml"))

def main(hparams={}):
    config = TRLConfig.update(default_config, hparams)

    val_split = 16
    dataset = load_dataset("ChaiML/dalio_scored_responses_v1")["train"]
    valid_ds = dataset[:val_split]
    train_ds = dataset[val_split:]

    eval_prompts = valid_ds["input_text"]

    dataset = (
        list(zip(
            train_ds["input_text"],
            train_ds["output_text"],
        )),
        train_ds["score"]
    )

    trlx.train(
        dataset=dataset,
        config=config,
        eval_prompts=eval_prompts,
    )


if __name__ == "__main__":
    main()
```

examples/ilql_sentiments.py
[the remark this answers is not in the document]
reciprocated: Yes, it's for consistency with other examples

trlx/orchestrator/offline_orchestrator.py
```diff
@@ -10,67 +53,84 @@ class OfflineOrchestrator(Orchestrator):
...
+        all_dones = []
+        for sample in samples:
+            length = 0
+            all_input_ids.append(torch.tensor(sum(sample, [])))
+            isoutput = False
+            actions_ixs, states_ixs = [], []
+            for phrase in sample:
+                if isoutput:
+                    actions_ixs.append(
+                        torch.arange(length - 1, length + len(phrase) - 1)
```
Dahoas, line 83: Why do states and actions append exactly the same thing?

trlx/orchestrator/offline_orchestrator.py
```diff
@@ -10,67 +53,84 @@ class OfflineOrchestrator(Orchestrator):
...
+        all_input_ids = []
+        all_actions_ixs = []
+        all_states_ixs = []
+        all_dones = []
+        for sample in samples:
+            length = 0
+            all_input_ids.append(torch.tensor(sum(sample, [])))
+            isoutput = False
+            actions_ixs, states_ixs = [], []
+            for phrase in sample:
```
Dahoas, line 80: So a sample can contain multiple "phrases" which we can reward?

Is there a reason we can't just separate each phrase into it's own datapoint with the "output" at the end(with all the prior dialog as context)? Do you see some advantage to treating things this way?

trlx/orchestrator/offline_orchestrator.py
```diff
@@ -1,7 +1,50 @@
...
 from trlx.orchestrator import Orchestrator, register_orchestrator
 from trlx.pipeline.offline_pipeline import ILQLRolloutStorage
+from trlx.utils import print_rank_0
+
+
+def tokenize_dialogue(  # noqa: C901
+    dialogue: Union[str, List[str]], tokenizer, max_length=2048, truncation_side="left"
+) -> List[int]:
+    """
+    Tokenize sample with the interleaved form of (question_1, answer_1, question_2, answer_2...)
```
Dahoas, line 15: Do we intend for answer_2 to depend on question_1, answer_1?

trlx/orchestrator/offline_orchestrator.py
```diff
@@ -1,7 +1,50 @@
...
+    if truncation_side == "left":
+        for phrase in reversed(dialogue):
+            tokens = tokenizer(phrase).input_ids[-ctx_length:]
+            ctx_length -= len(tokens)
+            out.insert(0, tokens)
+            if ctx_length == 0:
+                break
+
+        # in case of odd number of phrases (possibly due to truncation)
+        # since the first phrase always has to be a prompt, force it to be <bos>
```
Dahoas, line 34: Shouldn't the last phrase always be a prompt? Why must first always be prompt?

trlx/orchestrator/offline_orchestrator.py
```diff
@@ -1,7 +1,50 @@
...
+        if len(out) % 2 == 1:
+            if sum(map(len, out)) == max_length:
+                out[0].pop(0)
+            out.insert(0, [tokenizer.bos_token_id])
+
+    elif truncation_side == "right":
```
Dahoas, line 40: When would we want this? If we truncate on the right the agent is missing the most recent dialog

Dahoas:
Left some comments. In general I'm kinda skeptical about the need to support multiple "actions" (responses) in one trajectory. I don't see why we can't handle this just by splitting a dialog with 10 interactions into 10 samples, each of only has one action (at the very end of the dialog)

(Sorry if I misunderstood something)

trlx/orchestrator/offline_orchestrator.py
Dahoas: Perhaps including it is fine but I think it should be optional (and not the default)

reciprocated:
No Alex, this is a great question. The difference would be when you don't have intermediate rewards for each cumulative interaction `(q0, a0, r0)`, `(q0, a0, q1, a1, r1)` etc, but instead a single rating for the whole trajectory. Therefore it sense makes to credit assign the return to those preceding interactions as well and not necessarily only to the latest one, given that all interactions were in fact judged as a whole. And it's already totally optional depending how you pre-split your data yourself either way could work, it's to be experimentally determined which one is better

trlx/orchestrator/offline_orchestrator.py
reciprocated: This function accepts both dialogues and monologues and it's in anticipation of
- https://github.com/CarperAI/trlx/issues/177

trlx/orchestrator/offline_orchestrator.py
reciprocated: Counting from the left the last one should be action, otherwise there would be nothing to prompt and it would be redundant since the trajectory (more specifically actions) might as well be graded without it. At least datasets we are interested in, to my knowledge, are of this form

trlx/orchestrator/offline_orchestrator.py
reciprocated: It can be refactored, the difference between the two is few lines below

trlx/orchestrator/offline_orchestrator.py
reciprocated: Yes, ILQL was designed to do this

trlx/trainer/nn/ilql_models.py
```diff
@@ -48,18 +51,20 @@ def heads(self, hidden_size: int, vocab_size: int):
 
     def loss(self, outputs, labels: ILQLBatch):
         logits, (qs, target_qs, vs) = outputs
+        terminal_mask = labels.dones[:, :-1]
```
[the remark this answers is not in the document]
reciprocated: It's the same as `attention_mask` except only accounting `states_ixs` (subset of tokens from which continuations were sampled), plus the last token for which is also masked. It lets `V(terminal) = 0` in `V[:, 1:] * terminal_mask[:, 1:]` and also masks padding in other losses

trlx/orchestrator/offline_orchestrator.py
reciprocated: Yes, both are supported and opt-in

Dahoas (approved):

PR #158, opened 2023-01-03
features/add release template

Mistobaan:
A draft of a release checklist to test out for the incoming v0.4.0


alan-cooney:
@Mistobaan how about automating this instead?

As an example - https://github.com/neelnanda-io/TransformerLens/blob/main/.github/workflows/release.yml

The way this works is you create a release (with a tag of e.g. v1.0.x) and click "generate" to automatically create the notes from merged PRs (since the last release). Then this kicks of an action that sets the correct version, runs any tests and then publishes to PyPi .

PR #160, opened 2023-01-03
Add Petals BLOOM-176b example

artek0chumak:
Add the distributed inference&training example of BLOOM-176b by [Petals](https://github.com/bigscience-workshop/petals). The example is placed in `examples/ppo_petals_sentiments.py`.

To execute the example, please install petals by `pip install git+https://github.com/bigscience-workshop/petals`.

~Also, I added changes in the core code~ All core code changes were republished in #161 
Wandb report can be found here: https://wandb.ai/artek-chumak/trlx/runs/1k6jqnfb?workspace=user-artek-chumak

PR #161, opened 2023-01-03
More flexible way to initialize models

artek0chumak:
This PR is a copy from https://github.com/CarperAI/trlx/pull/160

It contains internal changes in core codes:
- Provide base and reference models initializers
- The more flexible way to create heads on HydraModels:
  - add dtype in value head
  - find lm head if already have one

### DIFF
--- a/trlx/orchestrator/ppo_orchestrator.py
+++ b/trlx/orchestrator/ppo_orchestrator.py
@@ -27,7 +27,7 @@ class PPOOrchestrator(Orchestrator):
         reward_fn: Callable,
         metric_fn: Optional[Callable] = None,
         chunk_size: int = 512,
-        ref_model_provider: Callable = None,
+        model_provider: Callable = None,
     ):
         self.pipeline = pipeline
         self.trainer = trainer
@@ -39,8 +39,8 @@ class PPOOrchestrator(Orchestrator):
         self.pipeline_loader = self.trainer.accelerator.prepare(self.pipeline_loader)
         self.pipeline_iterator = iter(self.pipeline_loader)
 
-        if ref_model_provider is not None:
-            self.ref_model = ref_model_provider(self.trainer.config.model.model_path)
+        if model_provider is not None:
+            self.ref_model = model_provider(self.trainer.config, adapters=False)
         elif not hasattr(self.trainer.model, "frozen_head"):
             self.ref_model = self.trainer.get_arch(self.trainer.config)
 
--- a/trlx/orchestrator/ppo_orchestrator.py
+++ b/trlx/orchestrator/ppo_orchestrator.py
@@ -27,6 +27,7 @@ class PPOOrchestrator(Orchestrator):
         reward_fn: Callable,
         metric_fn: Optional[Callable] = None,
         chunk_size: int = 512,
+        ref_model_provider: Callable = None,
     ):
         self.pipeline = pipeline
         self.trainer = trainer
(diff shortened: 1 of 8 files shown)


trlx/trainer/nn/ppo_models.py
```diff
@@ -26,6 +26,10 @@
     whiten,
 )
 
+TMODEL_PROVIDER = Callable[
+    [transformers.PretrainedConfig], transformers.AutoModelForCausalLM
```
LouisCastricato, line 30: This won't work for seq2seq, or NeMo once integration there is finished.

trlx/trlx.py
```diff
@@ -16,6 +16,8 @@ def train(
     config: Optional[TRLConfig] = None,
     split_token: Optional[str] = None,
     logit_mask: Optional[List[List[bool]]] = None,
+    ref_model_provider: Optional[Callable] = None,
```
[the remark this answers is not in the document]
artek0chumak: Is it better to provide any additional parameters via TRLConfig?

trlx/trainer/nn/ppo_models.py
artek0chumak: Will changing to  just `Callable` be enough?

LouisCastricato: Yeah I think so.

trlx/trlx.py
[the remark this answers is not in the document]
LouisCastricato: @reciprocated I think this is your call.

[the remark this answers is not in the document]
reciprocated: I don't find problematic extending api with new arguments, however perhaps not necessarily through ambiguous `**kwargs` as in the line below

trlx/orchestrator/ppo_orchestrator.py
```diff
@@ -38,7 +39,9 @@ def __init__(
         self.pipeline_loader = self.trainer.accelerator.prepare(self.pipeline_loader)
         self.pipeline_iterator = iter(self.pipeline_loader)
 
-        if not hasattr(self.trainer.model, "frozen_head"):
+        if ref_model_provider is not None:
+            self.ref_model = ref_model_provider(self.trainer.config.model.model_path)
```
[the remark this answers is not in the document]
reciprocated: Also, it's interesting in principle, but when do you think `ref_model` should not be the same as `base_model`, to require two different function for instantiating

trlx/trainer/nn/ppo_models.py
```diff
@@ -327,9 +340,12 @@ def __init__(
                 lm_head=self.base_model.lm_head,
             )
         # Cache `transformer.forward` args for general use (avoids incompatible args across architectures)
-        self.base_model_transformer_args = inspect.getfullargspec(
-            self.base_model.transformer.forward
-        ).args
+        if base_model_transformer_args is not None:
+            self.base_model_transformer_args = base_model_transformer_args
```
reciprocated, line 344: @artek0chumak could you explain why this variable had to be introduced? Maybe as a more general way to control `__call__` method, we can get it done on the user-side by changing `base_model` instead

trlx/orchestrator/ppo_orchestrator.py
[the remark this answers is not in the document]
artek0chumak: base_model can contain adapters(like soft-prompts in Petals, that haven't to be initialized in the ref_model.

It's better to provide a function that can return the base or ref model based on its arguments. But this way can create unnecessary restrictions for the factory function.

trlx/trainer/nn/ppo_models.py
artek0chumak: That can be done because all transformers models have to use hf's interface. However, some of them(like Petals) cannot use all the passed inputs, like the attention masks.

I may remove this change and can resolve the issue on Petals' side. Will discuss this with my teammates.

LouisCastricato: Any resolution?

artek0chumak: Yes, Petals will be changed, so I will remove these changes.

PR #171, opened 2023-01-08
Fix default `delta_kwargs` handling

jon-tow:
* Makes `delta_kwargs` an optional config; previously it defaulted to an empty `dict` which breaks the delta handlers for runs that do not use delta-tuning. Discovered by @PhungVanDuy 

* Removes `gpt2` from the default `MODIFIED_MODULES_DICT` as its MLP layers are implemented as Conv1D layers which are not supported by the OpenDelta pacakge; see this line [here](https://github.com/thunlp/OpenDelta/blob/a25fb1b2aabb9a2ed55ca941a15b9321bb32377c/opendelta/delta_models/lora.py#L151) (Conv1D does not contain `in/out_features` attrs)

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -95,8 +95,8 @@ def test_hf_attr_getters(model_name: str):
     [
         "EleutherAI/gpt-j-6B",
         "EleutherAI/gpt-neox-20b",
-        "gpt2",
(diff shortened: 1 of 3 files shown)


LouisCastricato (approved):
LGTM

PR #172, opened 2023-01-09
Update generation utilities

reciprocated:
This PR will
- [x] enable sweeping over a single `gen_kwargs` value (e.g. `temperature`, `top_k`, `beta`) during periodic evaluations
- [x] change `reward_fn`'s signature from `reward_fn(samples)` to `reward_fn(samples, prompts, responses)`
- [x] add an optional stop keyword to `gen_kwargs` (e.g. Models indicate they have completed a response by generating a stop sequence, which is literally the string "Human:" – from Anthropic's 2021 "A General Language Assistant as a Laboratory for Alignment")
- [x] clean up evaluation result's outputs

Among HF's `generate` arguments `StoppingCriteria` can only stop all generations per batch, `eos_token_ids` accepts only `List[int]`. The other option is to customize `generate` method for every architecture or to trim samples after `generate` as was done here

https://wandb.ai/sorry/trlx/reports/-Update-generation-utilities-172---VmlldzozMzIxMjE1
https://wandb.ai/sorry/trlx/reports/Update-generation-utilities-172--VmlldzozMzIxMjE5

### DIFF
--- a/examples/architext.py
+++ b/examples/architext.py
@@ -6,7 +6,7 @@ import trlx
 from trlx.data.configs import TRLConfig
 
 
-def reward_fn(samples):
+def reward_fn(samples, **kwargs):
     "Gives a negative count of rooms for each sample"
     return [-sample.count(":") for sample in samples]
 
--- a/examples/ilql_sentiments.py
+++ b/examples/ilql_sentiments.py
@@ -29,7 +29,7 @@ def main(hparams={}):
         device=0 if int(os.environ.get("LOCAL_RANK", 0)) == 0 else -1,
     )
 
-    def metric_fn(samples: List[str]) -> Dict[str, List[float]]:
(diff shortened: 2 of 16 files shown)


LouisCastricato:
@PhungVanDuy Are we adding best-of-n here?

PhungVanDuy:
> @PhungVanDuy Are we adding best-of-n here?

Yes, we can add it here.

LouisCastricato:
Are we planning to merge tonight

reciprocated:
No, I still have to update every other example (currently grappling with T5's) and make regression plots. I will also add clarifying comments shortly so until don't rush with reviewing

jon-tow (changes requested):
Very nice update. I've left some comments for feedback. Could you also resolve the merge conflicts when you get a chance? Thanks 🙏

jon-tow (approved):

Issue #179, opened 2023-01-10
Update the read me to include how to save and load a model.


LouisCastricato:
### 📚 The doc issue

Currently it is unclear how to load and save models. It was answered here https://github.com/CarperAI/trlx/issues/169 by @PhungVanDuy. 

### Suggest a potential alternative/fix

We should update the readme to include this snippet that Duy wrote.

paulbricman:
Big up!

PR #180, opened 2023-01-10
Update README.md

LouisCastricato:
Add how to save files. See here. https://github.com/CarperAI/trlx/issues/179


PhungVanDuy:
I have added the save hf function already here: https://github.com/CarperAI/trlx/blob/main/trlx/trainer/accelerate_base_trainer.py#L202, you can mention this one in read me as well as save_best function and load hf.

LouisCastricato:
So its jus trainer.save?

PhungVanDuy:
> So its jus trainer.save?
Yes. It will save the hf format as well as the accelerate state.

PR #181, opened 2023-01-10
Fix HuggingFace `model.save_pretrained` for DDP

jon-tow:
* Ensures DDP models are unwrapped before accessing their underlying `base_model.save_pretrained` method.
* Moves the HF-base pre-trained save directory into the `directory` to avoid littering the root calling path, i.e. not adding an extra dir.
  * Aside: Are we sure we want to save the model twice here?

### DIFF
--- a/trlx/trainer/accelerate_base_trainer.py
+++ b/trlx/trainer/accelerate_base_trainer.py
@@ -198,11 +198,14 @@ class AccelerateRLTrainer(BaseRLTrainer):
 
     def save(self, directory: Optional[str] = None):
         """Creates a checkpoint of the optimizer, scheduler and model"""
-        directory = directory or self.config.train.checkpoint_dir
-        self.accelerator.save_state(directory)
-        self.accelerator.unwrap_model(self.model).base_model.save_pretrained(
-            f"{directory}/hf_model"
-        )
+        self.accelerator.save_state(directory or self.config.train.checkpoint_dir)
+
+    @abstractmethod
+    def save_pretrained(self, directory: Optional[str] = None):
+        """Save the model and its configuration file to a directory, so that it can be re-loaded with the
+        `transformers.PreTrainedModel.from_pretrained` method.
+        """
+        pass
 
     def load(self, directory=None):
         """Load checkpoint of optimizer, scheduler and a model"""
--- a/trlx/trainer/accelerate_base_trainer.py
+++ b/trlx/trainer/accelerate_base_trainer.py
@@ -4,7 +4,7 @@ import os
 import sys
 from abc import abstractmethod
 from time import time
-from typing import Dict, Sequence, Tuple, Union
+from typing import Dict, Optional, Sequence, Tuple, Union
 
 import torch
 import torch.nn.functional as F
@@ -196,15 +196,13 @@ class AccelerateRLTrainer(BaseRLTrainer):
                 input_ids=input_ids, attention_mask=attention_mask, **kwargs
             )
 
-    def save(self, directory=None):
-        """Creates checkpoint of optimizer, scheduler and a model"""
-        self.accelerator.save_state(directory or self.config.train.checkpoint_dir)
-        if directory:
-            self.model.base_model.save_pretrained(f"hf_model_{directory}")
-        else:
-            self.model.base_model.save_pretrained(
-                f"hf_model_{self.config.train.checkpoint_dir}"
-            )
+    def save(self, directory: Optional[str] = None):
+        """Creates a checkpoint of the optimizer, scheduler and model"""
+        directory = directory or self.config.train.checkpoint_dir
+        self.accelerator.save_state(directory)
+        self.accelerator.unwrap_model(self.model).base_model.save_pretrained(
+            f"{directory}/hf_model"
+        )
 
     def load(self, directory=None):
         """Load checkpoint of optimizer, scheduler and a model"""
--- a/trlx/trainer/accelerate_base_trainer.py
+++ b/trlx/trainer/accelerate_base_trainer.py
@@ -221,7 +221,7 @@ class AccelerateRLTrainer(BaseRLTrainer):
         prompts_sizes = []
         lst_prompts = []
         generate_time = time()
-        for prompts in tqdm(self.eval_dataloader, desc="Generating samples"):
+        for prompts in self.eval_dataloader:
             if isinstance(prompts, torch.Tensor):
                 samples = self.generate_eval(prompts)
             else:
--- a/README.md
+++ b/README.md
@@ -34,6 +34,11 @@ trainer = trlx.train('EleutherAI/gpt-j-6B', dataset=[('dolphins', 'geese'), (1.0
 trainer.generate(**tokenizer('Q: Who rules the world? A:', return_tensors='pt'), do_sample=True)
 ```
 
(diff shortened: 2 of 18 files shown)


trlx/trainer/accelerate_base_trainer.py
```diff
@@ -196,15 +196,13 @@ def generate_eval(self, input_ids, attention_mask=None, **kwargs):
...
-            self.model.base_model.save_pretrained(
-                f"hf_model_{self.config.train.checkpoint_dir}"
-            )
+    def save(self, directory: Optional[str] = None):
+        """Creates a checkpoint of the optimizer, scheduler and model"""
+        directory = directory or self.config.train.checkpoint_dir
+        self.accelerator.save_state(directory)
+        self.accelerator.unwrap_model(self.model).base_model.save_pretrained(
+            f"{directory}/hf_model"
+        )
```
reciprocated, line 205: How about adding a seperate method `save_pretrained` to the `trainer` instead, delegated to each model since for ILQL saving `base_model` is not enough and it has to be wrapper over `PreTrainedModel` to achieve excepted result

reciprocated (commented):
Also since this is related, `trainer.save` seems to hang with multiprocessing not sure exactly why (while it doesn't two commits prior, before T5 on https://github.com/CarperAI/trlx/tree/4bfe3390f4c5fa0ed8aa43b09acdf16a7230f2cb) with
```
[2023-01-10 10:16:12,521] [INFO] [logging.py:68:log_dist] [Rank 0] [Torch] Checkpoint pytorch_model is begin to save!
WARNING:torch.distributed.elastic.agent.server.api:Received 2 death signal, shutting down workers
WARNING:torch.distributed.elastic.multiprocessing.api:Sending process 1894649 closing signal SIGINT
WARNING:torch.distributed.elastic.multiprocessing.api:Sending process 1894650 closing signal SIGINT
WARNING:torch.distributed.elastic.multiprocessing.api:Unable to shutdown process 1894649 via 2, forcefully exitting via 9
WARNING:torch.distributed.elastic.multiprocessing.api:Unable to shutdown process 1894650 via 2, forcefully exitting via 9
(.env) reciprocate@ip-26-0-129-228:/fsx/reciprocate/trlx$ [E ProcessGroupNCCL.cpp:821] [Rank 1] Watchdog caught collective operation timeout: WorkNCCL(SeqNum=259, OpType=ALLREDUCE, Timeout(ms)=1800000) ran for 1805329 milliseconds befo
re timing out.
[E ProcessGroupNCCL.cpp:821] [Rank 0] Watchdog caught collective operation timeout: WorkNCCL(SeqNum=258, OpType=ALLREDUCE, Timeout(ms)=1800000) ran for 1805439 milliseconds before timing out.
[E ProcessGroupNCCL.cpp:456] Some NCCL operations have failed or timed out. Due to the asynchronous nature of CUDA kernels, subsequent GPU operations might run on corrupted/incomplete data.
[E ProcessGroupNCCL.cpp:461] To avoid data inconsistency, we are taking the entire process down.
terminate called after throwing an instance of 'std::runtime_error'
  what():  [Rank 1] Watchdog caught collective operation timeout: WorkNCCL(SeqNum=259, OpType=ALLREDUCE, Timeout(ms)=1800000) ran for 1805329 milliseconds before timing out.
wandb: Waiting for W&B process to finish... (failed 15). Press Control-C to abort syncing.
[E ProcessGroupNCCL.cpp:456] Some NCCL operations have failed or timed out. Due to the asynchronous nature of CUDA kernels, subsequent GPU operations might run on corrupted/incomplete data.
[E ProcessGroupNCCL.cpp:461] To avoid data inconsistency, we are taking the entire process down.
terminate called after throwing an instance of 'std::runtime_error'
  what():  [Rank 0] Watchdog caught collective operation timeout: WorkNCCL(SeqNum=258, OpType=ALLREDUCE, Timeout(ms)=1800000) ran for 1805439 milliseconds before timing out.
```

```
compute_environment: LOCAL_MACHINE
deepspeed_config:
  deepspeed_multinode_launcher: standard
  gradient_accumulation_steps: 1
  gradient_clipping: 1.0
  offload_optimizer_device: none
  offload_param_device: none
  zero3_init_flag: false
  zero_stage: 1
distributed_type: DEEPSPEED
downcast_bf16: false
fsdp_config: {}
main_process_port: 1234
main_training_function: main
mixed_precision: bf16
num_machines: 1
num_processes: 2
use_cpu: false
```

```bash
accelerate launch --num_processes 2 --config_file single-zero1.yaml examples/ppo_sentiments.py```

File: trlx/trainer/accelerate_base_trainer.py (hunk 1 above)
Comment (LouisCastricato) [reply trlx/trainer/accelerate_base_trainer.py:205]: Agreed. @jon-tow make sure to update the readme as well.

Comment (jon-tow) [reply trlx/trainer/accelerate_base_trainer.py:205]: @reciprocated If you wrap the ILQL model with `transformers.PreTrainedModel` (or inherit it) and run `save_pretrained`, you still don't get the custom `generate` method needed for inference. I'm not sure what to do here because there isn't a reason to have `save_pretrained` if the custom generation implementation is not compatible with HF loading. Any hints/suggestions?

Comment (LouisCastricato) [reply trlx/trainer/accelerate_base_trainer.py:205]: ILQL is kinda like GeDi in that regard. Doesn't GeDi already exist on the hub?

Comment (reciprocated) [reply trlx/trainer/accelerate_base_trainer.py:205]: This is what Aman had suggested I believe, ILQL model will inevitably be tied to this repo. We can do the changes necessary at a later point, maybe `save_pretrained` on the base trainer with a `raise NotImplementedError` would be good for now

Comment (jon-tow) [reply trlx/trainer/accelerate_base_trainer.py]: Done. @reciprocated thanks for the suggestions 👍👍

**Review by LouisCastricato — APPROVED** [review]

**Comment by LouisCastricato** [comment]
Will merge after confirmed T5 test.

**Comment by jon-tow** [comment]
> Will merge after confirmed T5 test.

After a quick `"google/flan-t5-large"` run on PPO sentiments, loading seems fine.
```sh
>>> model = transformers.AutoModelForSeq2SeqLM.from_pretrained("ckpts/hf_model")
>>> tok = transformers.AutoTokenizer.from_pretrained("ckpts/hf_model")
Special tokens have been added in the vocabulary, make sure the associated word embeddings are fine-tuned or trained.
>>> inputs = tok(["hello, can you find where"], return_tensors='pt')
>>> tok.batch_decode(model.generate(**inputs), skip_special_tokens=True)
['I am?']
>>> model.config._name_or_path
'ckpts/hf_model'
>>> model.config.architectures
['T5ForConditionalGeneration']
```

LouisCastricato:
Great, merging.

paulbricman:
I was also running into errors on checkpoint save (before this latest commit). I pulled the latest and will update this comment to also confirm it works.

paulbricman:
@jon-tow Did you save your model using `trainer.save_pretrained` or `trainer.save`? They seem to populate the target folder with different things. @LouisCastricato this might be relevant for your recent README update?

I tried `trainer.save`, then running `zero_to_fp32.py` to obtain a `pytorch_model.bin`. Loading it with `transformers` does not error out, but it complains about having to initialize things, and most importantly, it only produces complete gibberish, compared to the generations I'm logging during training. Would this be likely to be caused by not using `trainer.save_pretrained` directly, and instead using the output of `trainer.save` and trying to make it work with HF?

Issue #16, opened 2022-10-06
Implement A2C


Dahoas:
Implement additional online RL algorithms

ML-Chen:
Hello, I'd like to be assigned to this issue!

geighz:
Hey @ML-Chen , I am wondering how you are doing, would you like to pair up and tackle this issue?

ML-Chen:
@geighz Would love to sometime later this week! What's your username in the CarperAI Discord server?

manavgarg:
Hey guys, I am interested in this too. Possible to include me here as well?

LouisCastricato:
Is this activate?

ML-Chen:
Working on this now

shermansiu:
Same, collabing with @ML-Chen.

shermansiu:
(I'm Blitz on the Discord)

RobertKirk:
Note that if you have a sufficiently generic implementation of PPO, you have A2C for free: https://github.com/vwxyzjn/a2c_is_a_special_case_of_ppo

shermansiu:
Yes, we're aware of that. It's just a matter of making the configurations nice so that it's easily accessible to end-users. There are also other RL algorithms that can be added.

shermansiu:
(The discussions for this are currently ongoing in the #trlx channel on the CarperAI Discord.)

shermansiu:
https://github.com/CarperAI/trlx/pull/183

PR #183, opened 2023-01-11
Add A2C

shermansiu:
Given how A2C is a special case of PPO (Huang et al., 2022), adding A2C to trlX becomes a matter of adding a few dedicated A2C configurations, as opposed to implementing A2C from scratch.

Some minor refactoring was necessary to get everything to work (and comply with flake8's standards), but overall the changes are quite minimal.

[1] S. Huang, A. Kanervisto, A. Raffin, W. Wang, S. Ontañón, and R. F. J. Dossa, A2C is a special case of PPO. 2022. Retrieved from https://arxiv.org/pdf/2205.09123.pdf


shermansiu:
I added a2c_sentiments.py, as requested by Louis Castricato. Unfortunately, there is a lot of overlap with ppo_sentiments.py but adding relative imports in the examples folder might increase confusion.

shermansiu:
Related issue: #16.

LouisCastricato:
> I added a2c_sentiments.py, as requested by Louis Castricato. Unfortunately, there is a lot of overlap with ppo_sentiments.py but adding relative imports in the examples folder might increase confusion.

That's fine.

LouisCastricato:
Do you have compute access yet? Or colab? Can you run a2c sentiment and post the wandb here?

trlx/utils/__init__.py
```diff
@@ -68,42 +70,46 @@ class OptimizerName(str, Enum):
     ADAMW: str = "adamw"
     ADAM_8BIT_BNB: str = "adam_8bit_bnb"
     ADAMW_8BIT_BNB: str = "adamw_8bit_bnb"
+    RMSPROP: str = "rmsprop"
     SGD: str = "sgd"
 
 
-def get_optimizer_class(name: OptimizerName):
+torch_optimizers: Dict[str, type] = dict(
```
LouisCastricato, line 77: @jon-tow I remember you being somewhat against this?

LouisCastricato (approved):

LouisCastricato:
Do NOT merge until we verify on wandb runs.

shermansiu:
I don't have access to the compute cluster yet. I can run the script after I get home from the University: I have a research computer at home that has a GPU with more VRAM.

trlx/utils/__init__.py
jon-tow: This should be fine. We don't really care about much of the torch optimizers outside of the current ones. I'm not even sure we should supply RMSProp as it's never really used for optimizing transformer language models.

configs/a2c_gptj.yml
```diff
@@ -0,0 +1,55 @@
+train:
```
jon-tow, line 1: Remove the `a2c_gptj.yml` unless tested thoroughly. We don't want folks wasting their compute resources on a large-ish tune with untested hparams.

jon-tow (changes requested):
Thanks, @shermansiu! I've made some minor change requests. Let's get some reports on this before doing a final review for merging. If you're unable to get resources soon-ish we can run things for you later in the week 👍

jon-tow:
> I don't have access to the compute cluster yet. I can run the script after I get home from the University: I have a research computer at home that has a GPU with more VRAM.

You should be able to get a reasonable signal + sanity checks from the `gpt2-imdb` model which is rather small. Let us know it goes!

trlx/utils/__init__.py
jon-tow: This should be fine. We don't really care about much of the torch optimizers outside of the current ones. I'm not even sure we should supply RMSProp as it's never really used for optimizing transformer language models.

LouisCastricato:
If you can get us run in the next 12 hours we can merge this for 0.4

shermansiu:
I made a run of A2C: unfortunately, the mean reward is quite volatile compared to that of PPO. I no longer think A2C is a good candidate algorithm for RLHF.

https://api.wandb.ai/report/shermansiu/lydosf6p

shermansiu:
This is even when I slash the learning rate from 1e-5 to 1e-7... Both are equally volatile, but 1e-7 has a lower mean reward.

RobertKirk:
> I no longer think A2C is a good candidate algorithm for RLHF.

It is the algorithm that DeepMind is using for all their RLHF work, and it seems to work well there. Probably there's some other hyperparameters/etc. that make it work there. They also use the KL penalty as an auxiliary loss rather than reward, which makes the RL reward simpler, which is maybe why A2C works for them.

shermansiu:
This is the relevant paper. Muesli is a higher priority item, but I could look at this later.
https://arxiv.org/pdf/1706.03741.pdf

conceptofmind:
> This is even when I slash the learning rate from 1e-5 to 1e-7... Both are equally volatile, but 1e-7 has a lower mean reward.

This is what was used in the Sparrow paper for A2C: 

"We extend the RL scheme of Menick et al. (2022); Perez et al. (2022), training a 70B A2C policy using Adafactor (Shazeer and Stern, 2018), a learning rate of 2 × 10−6, an effective batch size of 16, and 𝑙2-norm gradient clipping to a max norm of 1.0. Instead of the typical entropy term, we regularise by adding the KL divergence between the RL policy and the initial language model (SFT or Chinchilla) to the loss, with a weight 0.2. To reduce memory usage, we freeze the first 80% of the weights (64/80 transformer layers) to the pretrained values, share parameters between policy and value functions, and train with reduced precision using bfloat16 as in Rae et al. (2021) and stochastic rounding (Gupta et al., 2015). The value function predicts the final reward (without discounting) at each token. We implement the value function as an MLP with two hidden layers of size 2048, which takes as input the final transformer representation at each time step. We shard the models across 64 TPU v3 machines (Shoeybi et al., 2019)".

Issue #177, opened 2023-01-09
Add tokenizer truncation side option


jon-tow:
### 🚀 The feature, motivation, and pitch

Some tasks may benefit from truncating on different sides of batched prompts. We should add a configuration option to handle this from a high level as discussed [here](https://github.com/CarperAI/trlx/pull/145). Pasting discussion for quick view:

<img width="737" alt="Screenshot 2023-01-09 at 16 28 43" src="https://user-images.githubusercontent.com/41410219/211411825-7b107841-ecff-4603-8ef1-6d6bbb851b4b.png">

CC @reciprocated and @PhungVanDuy just in case y'all would like to add any extra comments or context.

PhungVanDuy:
Thank you for this issue, should we also consider adding the padding side along with the truncation side?

PR #189, opened 2023-01-13
Add tokenizer config

reciprocated:
This PR closes 
- #177

https://wandb.ai/sorry/trlx/reports/Difference-truncation_side-left-right--VmlldzozMzMwNTk5

### DIFF
--- a/trlx/data/configs.py
+++ b/trlx/data/configs.py
@@ -26,9 +26,6 @@ class ModelConfig:
     :param model_path: Path or name of the model (local or on huggingface hub)
     :type model_path: str
 
-    :param tokenizer_path: Path or name of the tokenizer (local or on huggingface hub)
-    :type tokenizer_path: str
-
     :param model_arch_type: Type of model architecture. Either "causal" or "seq2seq"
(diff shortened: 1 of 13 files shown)


trlx/data/configs.py
```diff
@@ -60,6 +56,30 @@ def from_dict(cls, config: Dict[str, Any]):
...
+    :type tokenizer_path: str
+
+    :param padding_side: Padding side
+    :type padding_path: str
+
+    :param truncation_side: Truncation side
+    :type truncation_side: str
+    """
+
+    tokenizer_path: str
```
jon-tow, line 74: (Thinking out loud on usability): I feel like we could make this optional and default to using the `model_path` to load the tokenizer out of `AutoTokenizer.from_pretrained` in the base trainer)

jon-tow (approved):
Looks good to me! Great work as usual 🫡

trlx/data/configs.py
Mistobaan: Also the `tokenizer_` part seems redundant being already in a `TokenizerConfig `

PR #192, opened 2023-01-14
Make the config loading consistent across all example scripts.

shermansiu:
Rewriting the config loading like this ensures that the examples can be called from anywhere. Moreover, file streams are properly closed after the file contents are retrieved.


jon-tow (approved):
Thanks for cleaning this up! I've updated some of the more recent examples.

PR #193, opened 2023-01-14
 Clean up optimizer class loading.

shermansiu:
Flake8 was giving the C901 complaint, so I simplified the `get_optimizer_class` function. Moreover, now, adding a default PyTorch or a `bitsandbytes` optimizer is as simple as a one-liner.

### DIFF
--- a/trlx/utils/__init__.py
+++ b/trlx/utils/__init__.py
@@ -78,16 +78,6 @@ def get_distributed_config(accelerator: Accelerator):
     return dist_config
 
 
-class OptimizerName(str, Enum):
-    """Supported optimizer names"""
-
-    ADAM: str = "adam"
-    ADAMW: str = "adamw"
-    ADAM_8BIT_BNB: str = "adam_8bit_bnb"
-    ADAMW_8BIT_BNB: str = "adamw_8bit_bnb"
-    SGD: str = "sgd"
-
-
 torch_optimizers: Dict[str, type] = dict(
     inspect.getmembers(
         torch.optim, lambda x: inspect.isclass(x) and x.__name__ != "Optimizer"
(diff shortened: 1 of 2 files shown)


shermansiu:
@jon-tow

trlx/utils/__init__.py
```diff
@@ -86,39 +88,42 @@ class OptimizerName(str, Enum):
...
+            return optimizer_class
         except ImportError:
             raise ImportError(
-                "You must install the `bitsandbytes` package to use 8-bit AdamW. "
+                f"You must install the `bitsandbytes` package to use the {optimizer_name} optimizer. "
                 "Install with: `pip install bitsandbytes`"
             )
-    if name == OptimizerName.SGD.value:
-        return torch.optim.SGD
     supported_optimizers = [o.value for o in OptimizerName]
```
jon-tow, line 127: `supported_optimizers` needs to get updated now that `OptimizerName` doesn't account for the entire set of optimizers available from `torch.optim`.

trlx/utils/__init__.py
```diff
@@ -86,39 +88,42 @@ class OptimizerName(str, Enum):
...
 
-def get_optimizer_class(name: OptimizerName):
+torch_optimizers: Dict[str, type] = dict(
+    inspect.getmembers(
+        torch.optim, lambda x: inspect.isclass(x) and x.__name__ != "Optimizer"
+    )
+)
+torch_optimizers = {
+    name.lower(): optimizer for name, optimizer in torch_optimizers.items()
+}
```
jon-tow, line 98: This is a tad hacky but cool. If we do this then there's little point in having the OptimizerName enum other than explicitly laying out what optimizers are available.

jon-tow (changes requested):
Thanks for putting this together @shermansiu! I've left comments below. I'm hesitant to merge this as is since I don't mind that we're explicit in what optimizers are supported. Moreover, we only need ~3 of the `torch.optim` optimizers (Adam/AdamW/RAdam), so I'm not sure it makes sense to pull them all in.

LouisCastricato:
Imho don't merge unless it becomes heavily requested

trlx/utils/__init__.py
shermansiu: Fixed.

PR #194, opened 2023-01-15
Fix `summarize_rlhf` inference checkpoint paths

jon-tow:
* Fix inference paths to load the PPO-trained models instead of SFT checkpoints.


jon-tow:
Thanks, Duy! 🤗

PR #201, opened 2023-01-17
Make `Trainer.save_pretrained` sub-directory optional

jon-tow:
* Makes the `Trainer.save_pretrained` `/hf_model` sub-directory optional. Previously, if a `directory` was supplied to `save_pretrained` the `hf_model` name path would be hard joined as `{directory}/hf_model`. Now users can directly write to the path defined by `directory` instead.
  * Maybe the HF checkpoint dir should be an extra config? (Not sure if it's worth it).
* Corrects the repo `README.md` to properly document `save_pretrained` usage.


reciprocated (approved):
LGTM!

Issue #199, opened 2023-01-17
Add CD to publish PyPi Package


alan-cooney:
### 🚀 The feature, motivation, and pitch

It would be nice if we could `pip install trlx` in a project, rather than having to clone.

I suggest we use the approach that I also implemented for [TrasformerLens](https://github.com/neelnanda-io/TransformerLens):

- Use Poetry rather than SetupTools/PIP to manage dependencies when developing the project (end users can still use pip to install it). This automatically detects dependency clashes that can be a nightmare with PyPi releases. It also has some nice helper tooling (e.g. for bumping package versions). Downside is it's a bit slower.
- Break out checks from the pre-commit hook into a workflow that runs on all PRs and pushes to main (so that we can see which part failed and why) - things like pytest and mypi. Also run this on all common Python versions (a bit like https://github.com/neelnanda-io/TransformerLens/actions/runs/3934543895/jobs/6729387264 )
- Have a release workflow, so that when you create a release (e.g. with v.1.x.0) on GitHub it automatically bundles up the package and releases to PyPi (if the checks are still passing)

### Alternatives

Do all the above but use SetupTools still rather than Poetry. Advantage is it's still more known (although Poetry is very popular now), disadvantage is a lot of stuff is harder (e.g. bumping package versions in the CD automatically) and it doesn't check for conflicts.

### Additional context

Happy to implement this

jon-tow:
Hi @alan-cooney! This is definitely something we'd like. I believe most of us are familiar with SetupTools so I'm partial to the alternative you've outlined.

> Break out checks from the pre-commit hook into a workflow that runs on all PRs and pushes to main (so that we can see which part failed and why) - things like pytest and mypi.
* What's different in this scenario from what's currently being done for PRs in the current workflow (e.g. see [Actions](https://github.com/CarperAI/trlx/actions))?

alan-cooney:
Ah yes I assumed that the build action was something else - we can just reference this directly for the checks part (and probably rename it to checks).

I'll create the PR

jovany-wang:
Was this done? Otherwise I'm willing to take it.

alan-cooney:
Sorry it hasn't been done yet. 

My personal opinion is that https://github.com/neelnanda-io/TransformerLens/blob/main/.github/workflows/release.yml represents best practice on this as a starting point (needs to be switched to use pip/setup tools though

PR #206, opened 2023-01-19
Create publish_to_pip.yml

BirgerMoell:
Made a build script for publishing the package to python. The script assumes that there is a secrets.PYPI_KEY that contains your key for publishing to pip.

### DIFF
--- a/.github/workflows/publish_to_pip.yml
+++ b/.github/workflows/publish_to_pip.yml
@@ -1,0 +1,35 @@
+name: Run tests and push to pypi
+on: [push]
+jobs:
+  build-and-publish:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@master
+    - name: Set up Python 3.7
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.7
+    - name: Install pypa/build
+      run: >-
+        python -m
+        pip install
+        build
+        --user
+    - name: Build a binary wheel and a source tarball
+      run: >-
+        python -m
+        build
+        --sdist
+        --wheel
+        --outdir dist/
+        .
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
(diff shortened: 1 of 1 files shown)


alan-cooney:
This relates to https://github.com/CarperAI/trlx/issues/199

@BirgerMoell the way I was going to set this up (in a PR) was so that the maintainers can simply create a GitHub release (with the desired version) and then the action can publish to PyPi with that version. You can see an example of this at https://github.com/neelnanda-io/TransformerLens/blob/main/.github/workflows/release.yml (where it also calls the checks workflow before doing this). What do you think about adding this in as well?

BirgerMoell:
I updated the code, I'm not 100% sure if that would automatically run when you publish since I haven't run github actions on publish before, but it would be a minor change to get it to work.

If you set up an account on pypi maybe you could try it out? @alan-cooney

alan-cooney:
I'm not a maintainer so I can't try directly I'm afraid - just making the suggestion above as I was going to pull this off with issue #199. You could however test this on a fork using your own test pypi account - https://test.pypi.org/

The action here wouldn't work exactly - as it doesn't set the version based on the GitHub release version. It also doesn't run the test workflow.

For what it's worth I think you can just copy https://github.com/neelnanda-io/TransformerLens/blob/main/.github/workflows/release.yml exactly and then:

- Replace poetry with pip
- Find something that set the version programatically (to replace `poetry version x`)
- Link to the checks workflow here to run that first

Happy to help do this if you want btw - I just don't use pip anymore as it creates too many headaches with package dependency conflicts when maintaining libs (because it doesn't check for them).

cat-state:
Thanks for this PR! I'll try this out now. One issue with the current packaging (if you `pip install git+github.com...`) is that it doesn't include the examples or configs, but all our examples use them as a base. This might make it not possible to follow the examples when using trlx as a library, so first we should probably inline the configs as classes into the library

PR #209, opened 2023-01-22
Enable training with Tensorboard tracking

marcobellagente93:
Currently only wandb logging is supported, and using tensorboard will result in a number of small blockers.

This small PR removes the following blockers:
- a logging_dir is added to the training config to be used by trackers needing a local folder (e.g. tensorboard)
- an assertion is places before calling `init_trackers` to check the specified tracker
- a flattened config is used for Tensorboard logging
- **Optional** slight renaming (trackers -> tracker) and interface change assuming a single tracker is used

**Disclaimers**
- Tensorboard logging can't take nested dict. the provided flattened dict removes one level to log **most** but not all hyperparameters specified in the config
- There seems to be some weird behavior in Tensorboard with hps. They can be downloaded (with the correct values) but not displayed, see also: https://github.com/tensorflow/tensorboard/issues/5476 

CC @daia99


daia99:
We may want to remove the tensorboard files for the PR!

marcobellagente93:
> We may want to remove the tensorboard files for the PR!

trlx/trainer/accelerate_base_trainer.py
```diff
@@ -78,19 +78,32 @@ def __init__(self, config, **kwargs):
             dist_config = get_distributed_config(self.accelerator)
             config_dict["distributed"] = dist_config
             init_trackers_kwargs = {}
-            if "wandb" in config.train.trackers:
+            # HACK: Tensorboard doesn't like nested dict as hyperparams
+            config_dict_flat = {a:b for (k,v) in config_dict.items() for (a,b) in v.items() if not isinstance(b, dict)}
```
cat-state, line 82: You can use trlx.utils.modeling.flatten_dict here

marcobellagente93: thanks for the suggestion! I replaced the dict comprehension with a call to flatten_dict(), since tensorboard also doesn't like lists I added a couple lines to split the optimizer betas

trlx/trainer/accelerate_base_trainer.py
```diff
@@ -78,19 +81,39 @@ def __init__(self, config, **kwargs):
...
-                config=config_dict,
-                init_kwargs=init_trackers_kwargs,
-            )
+
+                self.accelerator.init_trackers(
+                    project_name=self.config.train.project_name,
+                    config=config_dict,
+                    init_kwargs=init_trackers_kwargs,
+                )
+            else:
```
cat-state, line 104: Thanks!
Could you add back a comment explaining what this branch is for and the flattening?

marcobellagente93: Sure! 
- I had run into problems running trlx without wandb (I don't have an account as of now), and found an already opened issue precisely on this. The branch has minor modifications (which don't change the previous interface for wandb users) to allow tensorboard tracking
- The only tricky part is that wandb is pretty fancy and take nested dicts as logging params, this is not the case for tensorboard, hence the experiment config is fully flattened, and the only list is simply split apart (for the same reason)
Do let me know if anything is not clear or if I should add comments in the tensorboard specific logging

cat-state: Oh sorry, by branch I meant the `else` branch - i.e a short comment like
`else: # tracker == 'tensorboard' ` and `# flatten config for tensorboard, flatten lists in hparams into flattened config`

trlx/trainer/accelerate_base_trainer.py
```diff
@@ -101,8 +101,10 @@ def __init__(self, config, **kwargs):
                     config=config_dict,
                     init_kwargs=init_trackers_kwargs,
                 )
-            else:
-                config_dict_flat = flatten_dict(config_dict)
+            else:  # only other suppoert tracker is tensorboard
```
cat-state, line 104: Typo - should be:
`else: # only other supported tracker is tensorboard`

cat-state (changes requested):

cat-state (approved):

cat-state:
Thanks for contributing! This LGTM now.

PR #210, opened 2023-01-22
Improve PPO readability

alan-cooney:
Adds mostly comments and renames a few variables to be more descriptive. Also deletes two unused methods.

I've run an example with wandb to check this doesn't change anything unexpectedly - https://wandb.ai/alancooney/trlx/runs/aew4nu69 (same as before the changes)

Co-authored-by: @jezgillen


alan-cooney:
@reciprocated to help with documenting this code, are you able to confirm what `for _ in range(self.n_updates_per_batch):` is intended to do in the base trainer?

https://github.com/CarperAI/trlx/blob/84a07113d8ced1497c5a7941a02f9c534bb4e185/trlx/trainer/accelerate_base_trainer.py#L445-L449

Git blame suggests this is your code. My understanding is it runs the same batch multiple times - is that as expected? That is to say it lets the user repeat a batch, rather than e.g. having multiple steps e.g. turns in a text-based game (which I don't think the library supports).

reciprocated:
Hey there! `n_updates_per_batch` let's perform multiple gradient updates on the same batch of data, this is a common PPO implementation detail. Here is an excerpt from PPO's paper abstract [1] as reference:
> Whereas standard policy gradient methods perform one gradient update per data sample, we propose a novel objective function that enables multiple epochs of minibatch updates

[1] https://arxiv.org/abs/1707.06347

alan-cooney:
> > Whereas standard policy gradient methods perform one gradient update per data sample, we propose a novel objective function that enables multiple epochs of minibatch updates

Awesome thanks for the quick reply! I'll add this in to the comments

trlx/orchestrator/ppo_orchestrator.py
```diff
@@ -92,7 +114,7 @@ def make_experience(self, num_rollouts: int = 1024, iter_count: int = 0):  # noq
                 )
                 for output in outputs
             ]
-            response_tensors = torch.vstack(outputs).to(device)
+            padded_samples = torch.vstack(outputs).to(device)
```
reciprocated, line 117: Probably `padded_outputs` (or rather `output_tensors` in view of `query_tensors`) is better suited name here, since by a sample `prompt + output` is implied everywhere else

trlx/orchestrator/ppo_orchestrator.py
```diff
@@ -178,74 +200,98 @@ def make_experience(self, num_rollouts: int = 1024, iter_count: int = 0):  # noq
...
-                logprobs = logprobs_from_logits(logits, all_tokens)
-                ref_logprobs = logprobs_from_logits(ref_logits, all_tokens)
+                logprobs = logprobs_of_labels(logits, all_tokens)
+                ref_logprobs = logprobs_of_labels(ref_logits, all_tokens)
 
-            n = samples.shape[0]
+            n_samples: int = samples.shape[0]
             logprobs = logprobs.cpu()
             ref_logprobs = ref_logprobs.cpu()
             query_tensors = query_tensors.cpu()
```
reciprocated, line 214: Might also rename `query` to `prompt` for consistency

trlx/trainer/accelerate_ppo_trainer.py
```diff
@@ -17,37 +18,70 @@
...
 
+        # Setup rollout logging
         if config.train.rollout_logging_dir is not None:
             self.log_rollouts = True
             self.setup_rollout_logging(config)
         else:
             self.log_rollouts = False
 
+        # Setup the rollout store
+        # Rollouts contain the query & response, log probs, values
```
reciprocated, line 56: prompt & output here for consistency again

reciprocated (changes requested):
Splendid, what's left is to merge changes from the main and do a pre-commit run and I've also left a small suggestion to remain consistent in naming throughout. Thanks for the effort @alan-cooney @jezgillen

LouisCastricato:
Let's get this merged

trlx/orchestrator/ppo_orchestrator.py
alan-cooney: I think `padded_outputs` makes sense - made the change

trlx/orchestrator/ppo_orchestrator.py
alan-cooney: Done

trlx/orchestrator/ppo_orchestrator.py
alan-cooney: How about `sample_outputs` to be even clearer?

alan-cooney:
Thanks for the review @reciprocated 

Note I've had to resolve a tonne of merge conflicts, so it's worth checking those are in line with expectations (they're primarily with your commits). 

Given the size of this PR please can you take a look as soon as possible & merge, otherwise we'll likely get other merge conflicts as well.

reciprocated (approved):
Marvellous, merging!

https://wandb.ai/sorry/trlx/reports/Improve-PPO-readability-210---VmlldzozNDg3ODEx

alan-cooney:
Thanks for the quick review!

PR #213, opened 2023-01-23
Make `make_head` accept dtype parameter

reciprocated:
This PR forces value heads to be the same `dtype` as the rest of the base model

https://wandb.ai/sorry/trlx/reports/Fix-heads-dtype--VmlldzozMzk3OTM3


cat-state (approved):
LGTM

PR #218, opened 2023-01-24
Move to exact kl divergence rather than estimate

alan-cooney:
This is computationally just as efficient


PhungVanDuy:
@alan-cooney  this looks great! Can you try to run on the summarization example?

alan-cooney:
@PhungVanDuy I'm only developing this on a 16GB GPU so can't run that example unfortunately.

Is the reviewer able to test this as well on a seq2seq model please, as I haven't tested that branch? Otherwise this PR should be good to go.

LouisCastricato:
We're getting you more GPUs.

trlx/orchestrator/ppo_orchestrator.py
```diff
@@ -193,29 +193,34 @@ def make_experience(self, num_rollouts: int = 1024, iter_count: int = 0):  # noq
...
                 ends = (response_tensors[:, start:] != 0).sum(1)
-                all_logprobs = [logprobs[ix, start : ends[ix]] for ix in range(n)]
-                all_values = [values[ix, start - 1 : ends[ix] - 1] for ix in range(n)]
+
+                # Calculate the KL Divergence
+                logprobs_all_vocab = F.log_softmax(
+                    logits[:, start:-1, :], dim=-1
+                )  # [sample x tokens x vocab]
+                ref_logprobs_all_vocab = F.log_softmax(
+                    ref_logits[:, start - 1 : -2, :], dim=-1
```
Dahoas, line 206: Why `ref_logits[:, start - 1 : -2, :]`?

Dahoas:
I want to point out that the previous method of computing the KL penalty term only took the difference between the logprobs at [tokens in the generated sequence](https://github.com/CarperAI/trlx/blob/2f230017a74448e16939a39f5e198cc8aa2eece9/trlx/utils/modeling.py#L210) whereas now we are proposing to penalize the entire token distribution at each step of generation, likely resulting in a harsher penalty.

marcobellagente93:
I am running now the suggested example (https://github.com/CarperAI/trlx/pull/218#issuecomment-1403168899) from main, I'll do the same on this branch and post some wandb plot

trlx/orchestrator/ppo_orchestrator.py
```diff
@@ -193,29 +193,34 @@ def make_experience(self, num_rollouts: int = 1024, iter_count: int = 0):  # noq
             ref_logprobs = ref_logprobs.cpu()
             query_tensors = query_tensors.cpu()
             response_tensors = response_tensors.cpu()
+
             if self.trainer.config.model.model_arch_type == "seq2seq":
                 start = 1  # skip the <s> token
                 ends = (response_tensors[:, start:] != 0).sum(1)
-                all_logprobs = [logprobs[ix, start : ends[ix]] for ix in range(n)]
-                all_values = [values[ix, start - 1 : ends[ix] - 1] for ix in range(n)]
+
```
marcobellagente93, line 200: This errors out as all_logprobs has been removed but is used regardless in PPORLElement

Dahoas:
Thinking about this a big more I recalled [Fine-tuning with human preferences](https://arxiv.org/pdf/1909.08593.pdf) explicitly only regularizes with $log(\frac{\pi (response | input) }{ \pi_{ref} (response | input)})$ (see page 3 eq 2).  So I'd feel more comfortable retaining both regularizers until we get a better sense of which gives best performance on a range of tasks

PR #219, opened 2023-01-25
Support nested updates in `merge`

cat-state:
Currently `TRLConfig.update` will fail if given something like `{ "optimizer": { "name": "adam_8bit_bnb" } }"` since it'll check if the `optimizer` key is in `updates`, but `merge` won't add it because the only keys updated are nested. `merge` will check the keys to in every nested dict with the root level keys of the update config, which is problematic when changing an attr that is common across many subconfigs (e.g `"name"`), so this way allows proper nested updates.

### DIFF
--- a/trlx/data/configs.py
+++ b/trlx/data/configs.py
@@ -9,8 +9,9 @@ from trlx.data.method_configs import MethodConfig, get_method
 def merge(base: Dict, update: Dict, updated: Set) -> Dict:
     "Recursively updates a nested dictionary with new values"
     for k, v in base.items():
-        if isinstance(v, dict):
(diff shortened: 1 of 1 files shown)


jon-tow (approved):
@cat-state nice catch 🙏  can repro here https://colab.research.google.com/drive/1RBLvwxh8KQlCjjo1QtkLMXMu4GfxcZoX?usp=sharing

cat-state:
> @cat-state nice catch pray can repro here https://colab.research.google.com/drive/1RBLvwxh8KQlCjjo1QtkLMXMu4GfxcZoX?usp=sharing

thanks for the repro and test!

PR #221, opened 2023-01-25
Fix typo reward normalize summarize

PhungVanDuy:


jon-tow (approved):
Must've slipped during our final push hehe LGTM 👍

PR #226, opened 2023-01-26
Make experiment tracking optional

jon-tow:
Re-introduces optional tracking after #209


cat-state:
Oops, I forgot about this use case when approving the mentioned PR

cat-state (approved):

PR #228, opened 2023-01-26
Fix tokenizer missing form config.to_dict()

alan-cooney:
Fixes the missing property, which was breaking some use cases where hyperparameters are being tuned with Ray

Also orders alphabetically.


jon-tow (approved):
LGTM 👍 👍

PR #230, opened 2023-01-27
read tokenizer path from config correctly

JustinAWei:
In the specific config referenced by summarize_rlhf, it seems like the `model:` block does not have tokenizer_path. There does seem to be one in `tokenizer:` however.


LouisCastricato:
@PhungVanDuy

jon-tow (approved):
Thanks! This was a recent-ish update and was never reflected in the example

Issue #187, opened 2023-01-13
Script freeze with DeepSpeed


OleksandrKorovii:
### 🐛 Describe the bug

I try to run example `accelerate launch examples/ppo_sentiments.py` and after 100 steps script freeze last line in logs: 

`[2023-01-13 15:55:38,937] [INFO] [logging.py:68:log_dist] [Rank 0] [Torch] Checkpoint pytorch_model is begin to save!`

I was waiting during 1 hour, but nothing happened. I saw 100% load on my gpus in `nvidia-smi`.
I used DeepSpeed. I've tried  stage 1 and 2 - same result.

**BUT without  _DeepSpeed_ everything works fine**

### Which trlX version are you using?
install from source

### Additional system and package information

Python version: 3.10
GPU: 2x RTX3090Ti
accelerate==0.15.0
bitsandbytes==0.36.0.post2
deepspeed==0.7.7
tokenizers==0.13.2
torch==1.12.1+cu116
transformers==4.25.1

LouisCastricato:
This issue was resolved yesterday. Please pull from main. Thanks.

PR #242, opened 2023-01-31
Fix deepspeed state saving under `save_best` condition

reciprocated:
Fix of https://github.com/CarperAI/trlx/issues/187, complete `stats` dictionary is only available on rank 0 and therefore previously all other processes were on the other side of this if statement https://github.com/CarperAI/trlx/blob/a02f806eb17288456e24c16f1e709d30d958b41b/trlx/trainer/accelerate_base_trainer.py#L511-L514

https://wandb.ai/sorry/trlx/runs/xv5mczq1/logs

(sidenote: these line breaks look dreadful, can we perhaps increase black's --line-length to something like 120?)

### DIFF
--- a/trlx/trainer/accelerate_base_trainer.py
+++ b/trlx/trainer/accelerate_base_trainer.py
@@ -503,16 +503,21 @@ class AccelerateRLTrainer(BaseRLTrainer):
                         results = self.evaluate()
                         stats.update(results)
 
-                        # FIXME: seems to not work with zero and barriers don't seem to help
-                        if (
-                            self.config.train.save_best
-                            and int(os.environ.get("DEEPSPEED_ZERO_STAGE", -1)) == -1
-                        ):
-                            if (
-                                "reward/mean" in stats
-                                and stats["reward/mean"] > best_reward
-                            ):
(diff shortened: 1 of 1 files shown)


trlx/trainer/accelerate_base_trainer.py
```diff
@@ -503,17 +503,31 @@ def learn(self):  # noqa: C901
...
+                            else:
+                                do_save = False
+                            do_save = torch.tensor(
+                                do_save, device=self.accelerator.device
+                            )
+                            torch.distributed.all_reduce(
+                                do_save, torch.distributed.ReduceOp.MAX
+                            )
+                            if do_save:
+                                best_path = f"{self.config.train.checkpoint_dir}/best_checkpoint"
```
jon-tow, line 526: Nice - forgot to change this when updating the behavior to checkpoint into the specified dir.

jon-tow (approved):
Looks good! Fixes the best save issues I had for the PPO summarization runs :)

Re black formatting: Yeah that's fine with me. We have flake8 set to max-line-length 127 so makes sense to push black up to 127

PR #250, opened 2023-02-01
[style] Increase black's line length

reciprocated:
This PR bumps black's allowed line width in order to clean up sometimes awkward line breaks accumulated so far, yielding -400 LoC total as a result.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,3 +5,6 @@ build-backend = "setuptools.build_meta"
 [tool.isort]
 multi_line_output = 3
 profile = "black"
+
+[tool.black]
+line-length = 120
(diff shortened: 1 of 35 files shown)


pyproject.toml
```diff
@@ -5,3 +5,6 @@ build-backend = "setuptools.build_meta"
 [tool.isort]
 multi_line_output = 3
 profile = "black"
+
```
jon-tow, line 8: To fix the `isort` issue add `trlx` to the source paths:

```toml
[tool.isort]
src_paths="trlx"
...
```

jon-tow (changes requested):
Nice! Let's fix the `isort` issue once and for all (see suggestion below) and get this merged.

Issue #239, opened 2023-01-31
Verbosity during Rollout + Eval stages in trlx training


JustinAWei:
### 🚀 The feature, motivation, and pitch

I may be missing something, but the eval and rollout stages are a black box to me. The only insight I get into what is happening during these phases is prints in `reward_fn`. This is especially problematic with larger sequence_len and num_rollouts.

I recently came across a particularly hard-to-debug issue where having anything but a trivial size for the eval set (during multi-gpu training) would cause GPUs to time out its `NCC AllGather` operation due to the reward model taking 30+ minutes on the eval set. 

I had no idea what was going on under the hood at all and eventually figured out the problem via ablations of config values and checking GPU power usage graphs. If there was some indication on the stage of the training (eval) or the operation (waiting on reward model for eval rewards), this would've been way easier. 

As of now, after understanding some of the code, my strategy for debugging has been to add prints in the source code and rebuild the library; however, it would be such a QoL thing to have information on the main phases of training for other who might want to debug things in this library without needing to read a lot of the source code.

JustinAWei:
I have thought about contributing the prints myself but I am not confident I understand the full library yet

jon-tow:
Agreed - we'll be looking into this ASAP

Related Issues:
- [ ] #178

jon-tow:
Initial verbosity support has been merged. See usage details [here](https://github.com/CarperAI/trlx#logging). Please let us know if there is any other information you'd like to see logged during training.

Thank you for raising this issue!

Issue #178, opened 2023-01-10
Fix progress bar in `BaseRLTrainer.evaluate` for multi-processing


jon-tow:
### 🐛 Describe the bug

The `tqdm` progress bar for `BaseRLTrainer.evaluate` introduced in the T5 commit broke the RayTune utils; see the relevant lines below:
https://github.com/CarperAI/trlx/blob/0c5246f64e5e0ecb5fb2de65d440b122c792caf8/trlx/trainer/accelerate_base_trainer.py#L217-L225

It has been removed in the preparation of the `v0.4.0` release to ensure no other unexpected errors occur but we'd like to re-introduce this (robustly) as it makes clear what is happening during evaluation in an otherwise long-running sampling process.

CC @PhungVanDuy for awareness.

### Traceback

```
ray.exceptions.RayTaskError(LiveError): ray::ImplicitFunc.train() (pid=1701167, ip=26.0.134.192, repr=main)
  File "/trlx/.env/lib/python3.8/site-packages/ray/tune/trainable/trainable.py", line 367, in train
    raise skipped from exception_cause(skipped)
  File "/trlx/.env/lib/python3.8/site-packages/ray/tune/trainable/function_trainable.py", line 335, in entrypoint
    return self._trainable_func(
  File "/trlx/.env/lib/python3.8/site-packages/ray/tune/trainable/function_trainable.py", line 652, in _trainable_func
    output = fn()
  File "/tmp/ayu/trlx/examples/ppo_sentiments.py", line 51, in main
    trlx.train(
  File "/tmp/ayu/trlx/trlx/trlx.py", line 107, in train
    trainer.learn()
  File "/tmp/ayu/trlx/trlx/trainer/accelerate_base_trainer.py", line 385, in learn
    results = self.evaluate()
  File "/tmp/ayu/trlx/trlx/trainer/accelerate_base_trainer.py", line 224, in evaluate
    for prompts in tqdm(self.eval_dataloader, desc="Generating samples"):
  File "/trlx/.env/lib/python3.8/site-packages/tqdm/rich.py", line 116, in __init__
    self._prog.__enter__()
  File "/trlx/.env/lib/python3.8/site-packages/rich/progress.py", line 1169, in __enter__
    self.start()
  File "/trlx/.env/lib/python3.8/site-packages/rich/progress.py", line 1160, in start
    self.live.start(refresh=True)
  File "/trlx/.env/lib/python3.8/site-packages/rich/live.py", line 113, in start
[...]

PR #254, opened 2023-02-02
Add multi-process logger utility for status monitoring

jon-tow:
This PR introduces a basic `logging.Logger` utility for status monitoring from the console with multi-process support. By default, the logger is set to the `INFO` level but verbosity can be controlled by setting the `TRLX_LOG_LEVEL` environment variable to one of the standard logging [level names](https://docs.python.org/3/library/logging.html#logging-levels). Check the `README.md` update for more. 

- Replaces prints with logger calls to allow for more control over when and how trlx messages are displayed to users. (`print_rank_0` is kept in utils as a debug tool)

- Adds logger info to indicate evaluation stages such as reward and metric function process during training #239 

- Re-introduces @PhungVanDuy 's  progress bar for evaluating batches of samples #178

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,12 +67,12 @@ python -m trlx.sweep --config configs/sweeps/ppo_sweep.yml examples/ppo_sentimen
 
 trlX uses the standard Python `logging` library to log training information to the console. The default logger is set to the `INFO` level, which means that `INFO`, `WARNING`, `ERROR`, and `CRITICAL` level messages will be printed to standard output.
 
-To change the log level, you can use one of the direct setters. For example, to set the log level to `WARNING` you can use:
+To change the log level directly, you can use the verbosity setter. For example, to set the log level to `WARNING` use:
 
 ```python
 import trlx
 
-trlx.logging.set_verbosity_warning()
+trlx.logging.set_verbosity(trlx.logging.WARNING)
 ```
 
 This will suppress `INFO` level messages, but still print `WARNING`, `ERROR`, and `CRITICAL` level messages.
--- a/README.md
+++ b/README.md
@@ -65,16 +65,39 @@ python -m trlx.sweep --config configs/sweeps/ppo_sweep.yml examples/ppo_sentimen
 
 ## Logging
 
-trlX uses the standard Python `logging` library to log training information to the console. The default logger is set to `INFO` level, which means that `INFO`, `WARNING`, `ERROR`, and `CRITICAL` level messages will be printed to standard output. To control the verbosity, you can set the `TRLX_LOG_LEVEL` environment variable to one of the standard logging [level names](https://docs.python.org/3/library/logging.html#logging-levels).
+trlX uses the standard Python `logging` library to log training information to the console. The default logger is set to the `INFO` level, which means that `INFO`, `WARNING`, `ERROR`, and `CRITICAL` level messages will be printed to standard output.
 
-The following example sets the log level to `WARNING` for trlX logging events which will suppress `INFO` level messages, but will still print `WARNING`, `ERROR`, and `CRITICAL` level messages.
+To change the log level, you can use one of the direct setters. For example, to set the log level to `WARNING` you can use:
 
+```python
+import trlx
+
+trlx.logging.set_verbosity_warning()
 ```
-export TRLX_LOG_LEVEL=WARNING
+
+This will suppress `INFO` level messages, but still print `WARNING`, `ERROR`, and `CRITICAL` level messages.
+
+You can also control logging verbosity by setting the `TRLX_VERBOSITY` environment variable to one of the standard logging [level names](https://docs.python.org/3/library/logging.html#logging-levels):
+
+* `CRITICAL` (`trlx.logging.CRITICAL`)
+* `ERROR` (`trlx.logging.ERROR`)
+* `WARNING` (`trlx.logging.WARNING`)
+* `INFO` (`trlx.logging.INFO`)
+* `DEBUG` (`trlx.logging.DEBUG`)
+
+```sh
+export TRLX_VERBOSITY=WARNING
 ```
 
-> 💡 Tip: To reduce the amount of logging output, you might find it helpful to change log levels of third-party libraries used by trlX.
-For example, try adding `transformers.logging.set_verbosity_error()` to the top of your trlX scripts to suppress verbose messages from the `transformers` library (see their [logging docs](https://huggingface.co/docs/transformers/main_classes/logging#logging) for more details).
+By default, [`tqdm`](https://tqdm.github.io/docs/tqdm/) progress bars are used to display training progress. You can disable them by calling `trlx.logging.disable_progress_bar()`, otherwise `trlx.logging.enable_progress_bar()` to enable.
+
+Messages can be formatted with greater detail by setting `trlx.logging.enable_explicit_format()`. This will inject call-site information into each log which may be helpful for debugging.
+
+```sh
+[2023-01-01 05:00:00,000] [INFO] [ppo_orchestrator.py:63:make_experience] [RANK 0] Message...
+```
+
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -63,6 +63,15 @@ accelerate launch examples/simulacra.py
...
 ```
 
+## Logging
+
+trlX uses the standard Python `logging` library to log training information to the console. The default logger is set to `INFO` level, which means that `INFO`, `WARNING`, `ERROR`, and `CRITICAL` level messages will be printed to standard output. To control the verbosity, you can set the `TRLX_LOG_LEVEL` environment variable to one of the standard logging [level names](https://docs.python.org/3/library/logging.html#logging-levels).
+
+The following example sets the log level to `DEBUG` and prints all messages from trlX to the console.
+
+```
+export TRLX_LOG_LEVEL=DEBUG
```
reciprocated, line 73: Since there aren't any `logger.debug`s yet, what kind of messages should fall under this category in the future?

trlx/orchestrator/ppo_orchestrator.py
```diff
@@ -229,6 +242,8 @@ def make_experience(self, num_rollouts: int = 1024, iter_count: int = 0):  # noq
             ]
             ppo_rl_elements += new_ppo_rl_elements
             exp_time = clock.tick()
+            tbar.update(n if n < num_rollouts else num_rollouts)
```
reciprocated, line 245: Nit: `min(n, num_rollouts)`

trlx/orchestrator/ppo_orchestrator.py
```diff
@@ -55,9 +59,18 @@ def make_experience(self, num_rollouts: int = 1024, iter_count: int = 0):  # noq
         Takes `num_rollouts` prompts from `pipeline`, samples model and computes the
         KL againts a reference model. It then appends PPOElements to trainer's `store`
         """
+        logger.info("Collecting rollouts")
+        tbar = tqdm(
+            total=num_rollouts,
+            disable=not os.environ.get("RANK", 0) == "0",
```
reciprocated, line 65:
```suggestion
            disable=os.environ.get("RANK", "0") != "0",
```

trlx/orchestrator/ppo_orchestrator.py
```diff
@@ -55,9 +59,18 @@ def make_experience(self, num_rollouts: int = 1024, iter_count: int = 0):  # noq
...
+            position=0,
+            leave=False,
```
reciprocated, line 67: Why `leave` has to be false here? seems like colab doesn't particularly like it https://stackoverflow.com/a/57072638

reciprocated (commented):
Thanks Jon this will help with debugging timeouts massively! Left a few tiny nitpicks 🙏

README.md
jon-tow: Oh right! It's probably better to update this example to showcase `WARNING`, assuming we'll have some warnings in the future about, say, deprecations in the API. I can't really think of any use cases for `logger.debug` right now 😅

trlx/orchestrator/ppo_orchestrator.py
jon-tow: I thought leaving the progress bars would be too verbose but they show up in the stdout logs anyway. Updated!

README.md
```diff
@@ -63,6 +63,17 @@ accelerate launch examples/simulacra.py
...
 ```
 
+## Logging
+
+trlX uses the standard Python `logging` library to log training information to the console. The default logger is set to `INFO` level, which means that `INFO`, `WARNING`, `ERROR`, and `CRITICAL` level messages will be printed to standard output. To control the verbosity, you can set the `TRLX_LOG_LEVEL` environment variable to one of the standard logging [level names](https://docs.python.org/3/library/logging.html#logging-levels).
+
+The following example sets the log level to `WARNING` for trlX logging events.
+
+```
+export TRLX_LOG_LEVEL=WARNING
```
reciprocated, line 73: So this instruction would just silence `logging.info`s, maybe it's worth mentioning that more straightforwardly like you did few lines below?

README.md
```diff
@@ -63,6 +63,17 @@ accelerate launch examples/simulacra.py
...
+```
+
+> 💡 Tip: To reduce the amount of logging output, you might find it helpful to change log levels of third-party libraries. For the `transformers` library, try setting `transformers.logging.set_verbosity_error()`.
```
reciprocated, line 76: That's really handy, they've added some annoying logs in 4.25.1 (especially if you set prompts to `[<eos>]`). Can this method be used for trlx as well?

jon-tow: Yup! I should clarify this - I meant to suggest calling `transformers.logging.set_verbosity_error()` from a user's trlx script e.g. at the top of `examples/ilql_sentiments.py`.

jon-tow: (Let me know if it's any clearer 😄 )

reciprocated: Yes I understood, I will also clarify that I meant to ask wether you currently do `trlx.logging.set_verbosity_error()`

jon-tow: No, not currently. Maybe we should include logging methods in a similar way to transformers. What do you think?

reciprocated: I mean since we're already half-way there we might as well indulge, unless it's harder than it seems of course 🙂

jon-tow: 😂 will do!

jon-tow:
Update: I've added Hugging Face's logging API (used across their projects [`datasets`](https://github.com/huggingface/datasets/blob/2.9.0/src/datasets/utils/logging.py), [`diffusers`](https://github.com/huggingface/diffusers/blob/main/src/diffusers/utils/logging.py), [`transformers`](https://github.com/huggingface/transformers/blob/820c46a707ddd033975bc3b0549eea200e64c7da/src/transformers/utils/logging.py))

reciprocated (approved):
Nice, it looks really pretty especially when training with larger models!

(also I've observed that deepspeed's [logger](https://github.com/microsoft/DeepSpeed/blob/master/deepspeed/utils/logging.py) does everything similarly and can be filtered with `deepspeed.utils.logger.setLevel(logging.ERROR)`)

README.md
```diff
@@ -63,6 +63,41 @@ accelerate launch examples/simulacra.py
...
+
+```python
+import trlx
+
+trlx.logging.set_verbosity_warning()
+```
+
+This will suppress `INFO` level messages, but still print `WARNING`, `ERROR`, and `CRITICAL` level messages.
+
+You can also control logging verbosity by setting the `TRLX_VERBOSITY` environment variable to one of the standard logging [level names](https://docs.python.org/3/library/logging.html#logging-levels):
```
Mistobaan, line 80: is it `TRLX_VERBOSITY` or `TRLX_LOG_LEVEL` ?
I kind of like more `TRLX_LOG_LEVEL`

trlx/utils/logging.py
```diff
@@ -0,0 +1,352 @@
...
+            - `trlx.logging.WARNING`
+            - `trlx.logging.INFO`
+            - `trlx.logging.DEBUG`
+    """
+
+    _configure_library_root_logger()
+    _get_library_root_logger().setLevel(verbosity)
+
+
+def set_verbosity_info():
```
Mistobaan, line 173: so much code that does the same thing ... just give one option like the python library 

```python
trlx.logging.set_level(trlx.logging.INFO)
```

Mistobaan (changes requested):
- I would not use all this code for logging.
- push the rank configurations around the code that access global variables inside the logging configuration. As a rule of thumb the env variables should be access at the start of the process at configuration time and only once. 
- decide on the env variable to use TRLX_LOG_LEVEL vs TRLX_VERBOSITY
- printing tables in the logs is a nightmare for devops because parsing the output programmatically becomes very hard.

jon-tow:
@Mistobaan Yes, I agree with your points but the idea was to keep the API consistent with Hugging Face's logging API which is familiar to many. See links from https://github.com/CarperAI/trlx/pull/254#issuecomment-1416115419

If you have strong opinions about not doing this I can revert to the commit prior to introducing this (3034595ec86a9292de06cd3eb40972e24dd1b1bf).

Issue #238, opened 2023-01-30
PPO fp16


reciprocated:
### 🐛 Describe the bug

```
Traceback (most recent call last):
  File "examples/ppo_sentiments.py", line 61, in <module>
    main()
  File "examples/ppo_sentiments.py", line 52, in main
    trlx.train(
  File "/trlx/trlx/trlx.py", line 81, in train
    orch.make_experience(config.method.num_rollouts)
  File "/trlx/trlx/orchestrator/ppo_orchestrator.py", line 161, in make_experience
    logits, *_, values = self.trainer.model(
  File "/trlx/.env/lib/python3.8/site-packages/torch/nn/modules/module.py", line 1190, in _call_impl
    return forward_call(*input, **kwargs)
  File "/trlx/.env/lib/python3.8/site-packages/deepspeed/utils/nvtx.py", line 11, in wrapped_fn
    return func(*args, **kwargs)
  File "/trlx/.env/lib/python3.8/site-packages/deepspeed/runtime/engine.py", line 1836, in forward
    loss = self.module(*inputs, **kwargs)
  File "/trlx/.env/lib/python3.8/site-packages/torch/nn/modules/module.py", line 1190, in _call_impl
    return forward_call(*input, **kwargs)
  File "/trlx/trlx/trainer/nn/ppo_models.py", line 393, in forward
    transformer_outputs = self.base_model.transformer(**forward_kwargs)
  File "/trlx/.env/lib/python3.8/site-packages/torch/nn/modules/module.py", line 1190, in _call_impl
    return forward_call(*input, **kwargs)
  File "/trlx/.env/lib/python3.8/site-packages/transformers/models/gpt2/modeling_gpt2.py", line 832, in forward
    inputs_embeds = self.wte(input_ids)
  File "/trlx/.env/lib/python3.8/site-packages/torch/nn/modules/module.py", line 1190, in _call_impl
    return forward_call(*input, **kwargs)
  File "/trlx/.env/lib/python3.8/site-packages/torch/nn/modules/sparse.py", line 160, in forward
    return F.embedding(
  File "/trlx/.env/lib/python3.8/site-packages/torch/nn/functional.py", line 2210, in embedding
    return torch.embedding(weight, input, padding_idx, scale_grad_by_freq, sparse)
RuntimeError: Expected tensor for argument #1 'indices' to have one of the following scalar types: Long, Int; but got torch.cuda.HalfTensor instead (while checking arguments for embedding)
```
```
[...]

Issue #264, opened 2023-02-02
Cannot run PPO policy for summarization RLHF example


imraviagrawal:
### 🐛 Describe the bug

accelerate launch --config_file ./configs/default_accelerate_config.yaml trlx_gptj_text_summarization.py

Exception: Error: Trying to access a method that has not been registered

![Screen Shot 2023-02-02 at 11 01 57 AM](https://user-images.githubusercontent.com/24846591/216425273-f85474e4-0ef9-45ee-b864-6f652db515e7.png)

Using all of the default parameters as provided in the example. 

### Which trlX version are you using?

0.3.0

### Additional system and package information

pytorch=1.10.2+cu113, transformer=4.23.1

PhungVanDuy:
@imraviagrawal  can you try to run in the root folder like: `accelerate launch --config_file ./configs/default_accelerate_config.yaml examples/summarize_rlhf/trlx_gptj_text_summarization.py`?

imraviagrawal:
This worked, now I have another issue 

![image](https://user-images.githubusercontent.com/24846591/216684474-76a76a8b-f6a0-438f-8629-d76205630b51.png)

PhungVanDuy:
> This worked, now I have another issue, using all of the default parameters
> 
> ![image](https://user-images.githubusercontent.com/24846591/216684474-76a76a8b-f6a0-438f-8629-d76205630b51.png)

@imraviagrawal can you upgrade trlx with the newest version, I tried to re-run but cannot get this issue.

agassi4013:
> > This worked, now I have another issue, using all of the default parameters
> > ![image](https://user-images.githubusercontent.com/24846591/216684474-76a76a8b-f6a0-438f-8629-d76205630b51.png)
> 
> @imraviagrawal can you upgrade trlx with the newest version, I tried to re-run but cannot get this issue.

I run this example built yesterday, got same issue 

pytorch=1.10.0+cu113, transformer=4.26.0

trained sft and reward model based on gpt-neo-1.3B

PhungVanDuy:
> > > This worked, now I have another issue, using all of the default parameters
> > > ![image](https://user-images.githubusercontent.com/24846591/216684474-76a76a8b-f6a0-438f-8629-d76205630b51.png)
> > 
> > 
> > @imraviagrawal can you upgrade trlx with the newest version, I tried to re-run but cannot get this issue.
> 
> I run this example built yesterday, got same issue
> 
[...]

PR #279, opened 2023-02-06
[fix] Set deepspeed's fp16 `auto_cast` to false

reciprocated:
Fix of #238 

https://github.com/huggingface/accelerate/blob/4e5cc0c6b9c28fc569bfc70dd0508225d187fbf9/src/accelerate/utils/dataclasses.py#L527

https://wandb.ai/sorry/trlx/reports/-fix-Set-deepspeed-s-fp16-auto_cast-to-false-279--VmlldzozNDkzMzkx

Thanks @ZHAOTING for finding this fix

### DIFF
--- a/trlx/trainer/accelerate_base_trainer.py
+++ b/trlx/trainer/accelerate_base_trainer.py
@@ -43,10 +43,16 @@ class AccelerateRLTrainer(BaseRLTrainer):
     RL model trainer with an `accelerate` based backend
     """
 
-    def __init__(self, config, **kwargs):
+    def __init__(self, config, **kwargs):  # noqa: C901
         super().__init__(config, **kwargs)
         self.max_length = config.train.seq_length
         self.accelerator = Accelerator(log_with=config.train.tracker, logging_dir=config.train.logging_dir)
+
+        if self.accelerator.state.deepspeed_plugin is not None:
+            # by accelerate's default, arguments in `model.forward` would be casted to half
+            if "fp16" in self.accelerator.state.deepspeed_plugin.deepspeed_config:
+                self.accelerator.state.deepspeed_plugin.deepspeed_config["fp16"]["auto_cast"] = False
+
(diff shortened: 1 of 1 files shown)


cat-state:
Thanks @reciprocated and @ZHAOTING . Does this increa

se memory usage? Or are the model weights still being cast to fp16 somewhere else? I see in the report that the bf16 and fp16 runs have about the same but theres a run which seems to have more memory use:

2023-02-06-142143_738x134_scrot

reciprocated:
No, it doesn't increase the memory usage. For the baseline I took by accident a prior bf16 run, while on the fix is a run without accelerate, so the difference here is between bf16 and fp32 (it's not possible to make a fp16 run on the main). Just in case I made a few more runs with fp32, zero-fp32, zero-fp16, zero-bf16:
https://wandb.ai/sorry/trlx/reports/Set-deepspeed-s-fp16-auto_cast-to-false-279--VmlldzozNDk0OTMz

cat-state:
Ah okay, got it. This LGTM!

cat-state (approved):

reciprocated:
Hi! @xwwwwww, I'm able to reproduce the first part, it's related to the newest accelerate release which started to raise errors in this case, however I'm not getting the runtime error you're having. Maybe it's related to some particular versioning (at least that's what a quick google search suggests). Are you having the same problem for other examples as well?

xwwwwww:
Hi! @reciprocated 
For the second problem, I fix it as follows:

```
tmp = np.inf * torch.ones_like(xs)
min=torch.where(mask.bool(), xs, tmp).min(),
max=torch.where(mask.bool(), xs, -tmp).max(),
```

and my torch version is 1.10.1

PhungVanDuy:
> 

Can you try to upgrade your pytorch version? ref: https://github.com/CarperAI/trlx/issues/264#issuecomment-1422275089

PR #280, opened 2023-02-06
[feat] Add Accelerate SFT Trainer

reciprocated:
This PR gathers supervised finetuning from these two places[1][2] under the shared api

[1] https://github.com/CarperAI/trlx/blob/main/examples/summarize_rlhf/sft/train_gptj_summarize.py
[2] https://github.com/Dahoas/reward-modeling/blob/main/reward-modeling/finetune_base.py

https://wandb.ai/sorry/trlx/reports/Add-Accelerate-SFT-Trainer-280--VmlldzozNTAyNzQ1

### DIFF
--- a/trlx/trlx.py
+++ b/trlx/trlx.py
@@ -6,10 +6,11 @@ from trlx.utils import set_seed
 from trlx.utils.loading import get_orchestrator, get_pipeline, get_trainer
 
 
-def train(
+def train(  # noqa: C901
     model_path: Optional[str] = None,
     reward_fn: Optional[Callable[[List[str], List[str], List[str]], List[float]]] = None,
     dataset: Optional[Iterable[Tuple[str, float]]] = None,
+    samples: Optional[List[str]] = None,
     prompts: Optional[List[str]] = None,
     eval_prompts: Optional[List[str]] = None,
     metric_fn: Optional[Callable[[List[str], List[str], List[str]], Dict[str, List[float]]]] = None,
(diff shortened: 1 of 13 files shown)


cat-state:
Thanks @reciprocated , this mostly LGTM, could you update the other examples to also match the `samples, rewards` args instead of `dataset`?

reciprocated:
@cat-state, I've updated those, but perhaps I was overzealous with making `TRLConfig` a positional and required argument, since the only example that doesn't work is `simulacra.py` and by extension also simulacra's colab which git cloned main, since it didn't provide a config of its own. Should I maybe revert those changes? Not sure how much have you progressed with replacing the current configuration, and how much the above still matters

trlx/trlx.py
```diff
@@ -1,111 +1,117 @@
...
 
 from trlx.data.configs import TRLConfig
 from trlx.utils import set_seed
 from trlx.utils.loading import get_orchestrator, get_pipeline, get_trainer
 
 
-def train(
+def train(  # noqa: C901
+    config: Optional[TRLConfig] = None,
     model_path: Optional[str] = None,
```
cat-state, line 12: Maybe model path should still be the first arg? It's also whats still in the readme.

cat-state:
For the future config changes, we'll still need to keep the ModelConfig part around so keeping `model_path` makes sense for now. We can update the simulacra colab to give a model_path which should make this work?

trlx/trlx.py
reciprocated: yeah, reverted that bit

cat-state (approved):
okay! This LGTM now

PR #285, opened 2023-02-07
Add Colab Notebook for Sentiment

zswitten:
Code adapted from the [sentiment example](https://github.com/CarperAI/trlx/blob/main/examples/ilql_sentiments.py) and the [simulacra Colab](https://colab.research.google.com/drive/1vrmCLoHNlKvDVqJjMig-8tKDCfIEoym4?). One notable addition is the final cell which allows the user to see outputs from the RLHFed model. (Not sure why most of the tokens have a superfluous 'Ġ'.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,8 @@ pip install -e .
 For more usage see [examples](./examples). You can also try the colab notebooks below:
 | Description | Link |
 | ----------- | ----------- |
-| Simulacra Example | [![Open In Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/drive/1vrmCLoHNlKvDVqJjMig-8tKDCfIEoym4?usp=sharing)|
+| Simulacra (GPT2, ILQL) | [![Open In Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/drive/1vrmCLoHNlKvDVqJjMig-8tKDCfIEoym4?usp=sharing)|
+| Sentiment (GPT2, ILQL) | [![Open In Colab](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/drive/1lUGxD__nl3OeC0mHVAoew_PzMprRT_2w)
 
 ## How to Train
 


cat-state:
Thanks for contributing! I just looked at the notebook and it looks great mostly! 
One thing I thought when I tried it:
* You can set `config.train.tracker = None` to disable wandb, otherwise currently you must have a wandb account to try the notebook

cat-state (changes requested):
Use `trainer.tokenizer.decode(trainer_output)` instead of `convert_ids_to_tokens` and disable wandb.

zswitten:
Thank you for the review @cat-state! I made all the changes you suggested: 
- tracker disabled (no wandb)*
- no LOCAL_RANK check (device always set to 0)
- config.train.batch_size = 16
- Use trainer.tokenizer.decode(trainer_output)

I also made two other changes:
- Use `samples` and `rewards` instead of `dataset` in the `train` call as there was a warning that `dataset` is being deprecated
- Lower epochs to 10 to save people's compute credits.

I ran-all the notebook with standard GPU class and shape and it works e2e.

*Re: wandb, I copied that over from the Simulacra tutorial, so the owner of that Colab may want to change it.

cat-state (approved):

cat-state:
> Thank you for the review @cat-state! I made all the changes you suggested:
> 
>     * tracker disabled (no wandb)*
> 
>     * no LOCAL_RANK check (device always set to 0)
> 
>     * config.train.batch_size = 16
> 
>     * Use trainer.tokenizer.decode(trainer_output)
> 
> 
> I also made two other changes:
> 
>     * Use `samples` and `rewards` instead of `dataset` in the `train` call as there was a warning that `dataset` is being deprecated
> 
>     * Lower epochs to 10 to save people's compute credits.
> 
> 
> I ran-all the notebook with standard GPU class and shape and it works e2e.
> 
> *Re: wandb, I copied that over from the Simulacra tutorial, so the owner of that Colab may want to change it.

Thanks! This looks great now. Thanks for contributing!

Issue #253, opened 2023-02-02
Unexpected tokenization behavior of `PromptPipeline`


ZHAOTING:
### 🐛 Describe the bug

I am not sure whether this is intended or not, but in the `init()` of a `PromptPipeline`, prompt texts are converted to input id and attention mask tensors with a default value of `add_special_tokens=True` at [L20](https://github.com/CarperAI/trlx/blob/main/trlx/pipeline/offline_pipeline.py#L20).

For some tokenizers, it adds BOS and EOS tokens to the prompt, which might not be expected.
In my case, my policy model will generate nothing but another EOS token after seeing an EOS. This results in empty rewards and leads to [an endless loop when an orchestrator attempts to add new PPOElements].(https://github.com/CarperAI/trlx/blob/add-hh-example/trlx/orchestrator/ppo_orchestrator.py#L222).

### Which trlX version are you using?

Branch add-hh-example. At the [latest commit](https://github.com/CarperAI/trlx/commit/9745ec360ee43cafd3b63f213dbf04ca212e5e35).

LouisCastricato:
@jon-tow easy fix

cat-state:
Thanks for this issue, I just ran into this issue yesterday too.

cat-state:
@ZHAOTING it should be solved after merging #287 , let me know if it fixed it for your model

ZHAOTING:
@cat-state my experiment looks good at commit [b91da7b](b91da7b03d8e9fa0c0d6dce10a8f2611aca3013f).
Thanks for the fix!

PR #287, opened 2023-02-07
Set `add_special_tokens=False` to not add EOS unexpectedly

cat-state:
This PR addresses https://github.com/CarperAI/trlx/issues/253 . Set `add_special_tokens=False` and instead add BOS manually, matching what is done in the trainers.

comparison report with `ppo_sentiments`: https://wandb.ai/carperai/trlx/reports/PromptPipeline-Add-BOS---VmlldzozNTA2MTM4


cat-state:
I just realised this breaks the T5 example via 
```python
│ /home/a/trlx/examples/summarize_daily_cnn/t5_summarize_daily_cnn.py:26 in reward_fn              │
│                                                                                                  │
│   23 if __name__ == "__main__":                                                                  │
│   24 │                                                                                           │
│   25 │   def reward_fn(samples: List[str], prompts: List[str], outputs: List[str]):              │
│ ❱ 26 │   │   original_summaries = [prompt_label[prompt.strip()] for prompt in prompts]           │
│   27 │   │   scores = [                                                                          │
│   28 │   │   │   meteor.compute(predictions=[output.strip()], references=[original])["meteor"    │
│   29 │   │   │   for (original, output) in zip(original_summaries, outputs)                      │
│                                                                                                  │
│ /home/a/trlx/examples/summarize_daily_cnn/t5_summarize_daily_cnn.py:26 in <listcomp>             │
│                                                                                                  │
│   23 if __name__ == "__main__":                                                                  │
│   24 │                                                                                           │
│   25 │   def reward_fn(samples: List[str], prompts: List[str], outputs: List[str]):              │
│ ❱ 26 │   │   original_summaries = [prompt_label[prompt.strip()] for prompt in prompts]           │
│   27 │   │   scores = [                                                                          │
│   28 │   │   │   meteor.compute(predictions=[output.strip()], references=[original])["meteor"    │
│   29 │   │   │   for (original, output) in zip(original_summaries, outputs)                      │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
KeyError: 'Summarize: (CNN) -- This week marks one of the most-exciting non-major events of the golf season -- the Players Championship at the famed TPC Sawgrass. With 
a deep field and a great course, you won\'t want to miss any of the action. Before the tournament tees off, we had a chance to catch up with TPC Sawgrass PGA Head 
Professional Matt Borocz, who provided some inside insight on the home of the PGA Tour. PGA.com: Thanks for joining us. This week presents one of the most exciting on 
```

jon-tow:
>  Set add_special_tokens=False and instead add BOS manually, matching what is done in the trainers.

@cat-state wait where is this happening in the trainers? It looks like the `<>Trainer.tokenize()` methods that manually add BOS tokens never actually get called anymore, e.g. the base method  https://github.com/CarperAI/trlx/blob/e139ca4cd8b57d0e6a6a92af3df7a8f6386b6043/trlx/trainer/accelerate_ilql_trainer.py#L38-L52

Let me loop @reciprocated in here because I vaguely recall he brought this up before. Also, if these methods are dead we should probably remove them.

reciprocated:
Neither `AccelerateRLTrainer.tokenize`, `AccelerateILQLTrainer.tokenize` nor `NeMoILQLTrainer.tokenize` don't seem to be used anywhere at the moment. Actually there are only three places were tokenization happens: in both orchestrators and in the prompt pipeline (apart from examples)
https://github.com/CarperAI/trlx/blob/f115eeaa3cfd2c997a345b8891b5f9427f1a08ee/trlx/pipeline/offline_pipeline.py#L20
https://github.com/CarperAI/trlx/blob/f115eeaa3cfd2c997a345b8891b5f9427f1a08ee/trlx/orchestrator/offline_orchestrator.py#L32
https://github.com/CarperAI/trlx/blob/f115eeaa3cfd2c997a345b8891b5f9427f1a08ee/trlx/orchestrator/ppo_orchestrator.py#L117
Only the last needs to have `<eos>` appended (for the reward model & to make indexing work) and I think none really need `<bos>`

cat-state:
> are only three places were tokenization happens: in both orchestrators and in the prompt pipeline (apart from examples)

Oh right! so this was just dead code :sweat_smile: 
It seems like we still add EOS and BOS for ILQL in https://github.com/CarperAI/trlx/blob/f115eeaa3cfd2c997a345b8891b5f9427f1a08ee/trlx/orchestrator/offline_orchestrator.py#L23 
With the 20b or J or gpt2 the tokenizers don't add BOS/EOS by default it seems, however the original issue was when using a tokenizer that does

trlx/pipeline/offline_pipeline.py
```diff
@@ -15,13 +15,27 @@ class PromptPipeline(BasePipeline):
     Tokenizes prompts, unless they are already tokenized, and truncates them to `max_prompt_length` from the right
     """
 
-    def __init__(self, prompts: List[str], max_prompt_length: int, tokenizer=None):
+    def __init__(self, prompts: List[str], max_prompt_length: int, tokenizer: PreTrainedTokenizer):
         super().__init__()
-        model_inputs = tokenizer(prompts, truncation=True, padding=False, max_length=max_prompt_length)
-        prompts = model_inputs["input_ids"]
+
+        # manually prepend bos token if not already present to match RL trainers tokenization
```
jon-tow, line 21: This comment might be misleading since we no longer prepend BOS tokens from the trainers.

jon-tow (changes requested):
Thanks, @cat-state! This looks good 👍  Just leaving some questions for small edits.

trlx/pipeline/offline_pipeline.py
cat-state: Oh yeah, I'll change it. We still do add BOS and EOS for ILQL btw

trlx/pipeline/offline_pipeline.py
```diff
@@ -52,13 +54,28 @@ class PromptPipeline(BasePipeline):
...
+    def __init__(self, prompts: List[str], max_prompt_length: int, tokenizer: PreTrainedTokenizer):
         super().__init__()
-        model_inputs = tokenizer(prompts, truncation=True, padding=False, max_length=max_prompt_length)
-        prompts = model_inputs["input_ids"]
+
+        # manually prepend bos token if not already present to match `tokenize_dialog` for ILQL and
+        # default tokenizer behavior for PPO
+        if tokenizer.bos_token is not None:
+            prompts = [
+                tokenizer.bos_token + prompt if not prompt.startswith(tokenizer.bos_token) else prompt
```
reciprocated, line 64: Do we have to force adding BOS here? In ILQL it is only added in case there is only a single string passed, to make sure loss starts from 0-index of output (action_ixs[0] = 0) by prepending a "prompt" as BOS, and not when real `[prompt, output]` is passed (unless truncation comes into play). Don't want to change behaviour here as I'm finilizing HH PR with already shaby results as they are

reciprocated (commented):
Seems like in your report there are 5 different runs all with slightly different results, however all use `gpt2-imdb' for which this pr shouldn't change anything, is this only due to added BOS's influence?

jon-tow:
@reciprocated As an aside; could you clarify why you believe "none [of the tokenizations] really need `<bos>`"? Is this from empirical results you've noticed with RL? The "natural" thing to do with LMs is to include `<bos>` for inference.

cat-state:
> Seems like in your report there are 5 different runs all with slightly different results, however all use `gpt2-imdb` for which this pr shouldn't change anything, do you think this is only due to added BOS's influence?

The runs ran with the same seeds have the same results, I included 2 runs with each seed for both `base` and with bos to see how they change it

trlx/pipeline/offline_pipeline.py
cat-state: I see, that makes sense that the `bos` is only added to the the prompt for a completion if no prompt is given for that completion.

reciprocated:
> could you clarify why you believe "none [of the tokenizations] really need `<bos>`"? 

@jon-tow There I've meant that `<bos>` is really not needed algorithmically (unlike in the single output ILQL case), however I don't have any evidence nor counter-evidence that adding it is not advantageous in general. Yet there is such evidence for adding `<eos>`, something we actually don't do currently but I feel like ought to, from Anthropic's 2021 HHH paper:

> We also found that appending a special ‘end-of-context’ token to each sequence to unambiguously delineate the end of passage sometimes improves performance, as discussed in section C.4.

jon-tow:
@reciprocated  Yeah... I also haven't seen any experiments comparing results from `<bos>` vs no-`<bos>` for decoder models but folks make the argument you should because it gets you logprobs for the first token (however that may steer inference 🤷).
Thanks for linking the EOC stuff! (off-topic but I wonder if we did this for the HH RM training 🤔)

jon-tow (approved):
---
@cat-state Just noting that this subtly changes results for T5 models dailymail/cnn summarization example (PPO): https://api.wandb.ai/links/jon-tow/p3kg4ejf
Reward seems to improve better with this fix since removing the eos token before passing ids to `generate` is (probably?) the right thing to do.

reciprocated:
@jon-tow Yeah, but the first logprob of interest in PPO is in the beggining of output, not of the prompt, so unless `len(prompt) == 0` there is no reason to add `<bos>`. Regarding `<eos>` with HH RM, yes we did in https://github.com/Dahoas/reward-modeling/blob/deba81c2e9dab1514032800109325258caa470db/reward-modeling/rm_datasets.py#L93 @Dahoas, I also add it manually in: https://github.com/CarperAI/trlx/blob/34817b66fb9d18bab4f0a7a2676571f7de4e748a/examples/hh/ppo_hh.py#L36-L37

cat-state:
> improve better with this fix since removing the eos token before passing ids to `generate` is (probably?) the right thing to do.

Thanks for checking! Interesting, the differences seems less than inter-seed variance, so I guess there must be some effect

Issue #278, opened 2023-02-05
Remove orchestrator/trainer distinction


alan-cooney:
### 🚀 The feature, motivation, and pitch

As context, it looks like the trainer classes were split up into these two classes (trainer & orchestrator), as they had become very large. 

The way this has been done is to enforce instantiating both classes, and then they add themselves to each other as a property (i.e. `trainer.orchestrator` and `orchestrator.trainer`). Whilst quite a creative solution, this is unusual and has the following disadvantages:
- Prevents type checking from working properly (as it involves circular imports)
- Hard to read (as the resulting class is still huge, and now we just have to jump between two files instead)
- Hard to debug

Put simply, the abstraction doesn't work, so I think it makes most sense to merge them back together. We can then extract functionality where it does work (i.e. smaller discrete things like the kl divergence can be extracted), which will create a far more readable and robust codebase.

LouisCastricato:
Agreed. @jon-tow do you wanna tackle this?

jon-tow:
The orchestrators have been merged into their respective trainers; improving on the general abstraction should be easier now.

Thanks for raising this issue Alan!

PR #289, opened 2023-02-08
refactor: remove orchestrator abstraction from API

jon-tow:
This PR removes all of the orchestrator components for reasons outlined in #278.

Highlights for reviewer(s):

* Adds a new method to `AcceleratePPOTrainer` called `add_prompt_pipeline` that mimics the prompt pipeline loading and device placement of the removed PPO orchestrator. This is sort of awkward because it requires users to manually call the method before running `make_experience` (the same thing you have to do with `add_eval_pipeline`). Open to suggestions; I'd prefer to pass the pipeline to the trainer constructor but it breaks the config-first approach currently implemented (should discuss for future refactoring?).

* Removes dead code related to MagiCARP. Removes unused `utils.topk_mask`.

Reproduction reports:

* ILQL: https://api.wandb.ai/links/jon-tow/ygxa84nt
* PPO: https://api.wandb.ai/links/jon-tow/d4hurpqf


cat-state:
Thanks jon! I think you might have to also change something for the nemo trainer?

trlx/trlx.py
```diff
@@ -89,8 +89,12 @@ def train(  # noqa: C901
             eval_prompts = prompts[:batch_size]
 
         pipeline = get_pipeline(config.train.pipeline)(prompts, max_prompt_length, trainer.tokenizer)
-        orch = get_orchestrator(config.train.orchestrator)(trainer, pipeline, chunk_size=config.method.chunk_size)
-        orch.make_experience(config.method.num_rollouts)
+        trainer.add_prompt_pipeline(pipeline)
```
cat-state, line 92: Re `add_prompt_pipeline` - yeah that is a bit awkward, it should be possible to pass it via args if you move the `get_trainer` call into the PPO part of the branch?
You could also probably replace the `get_pipeline` with `PromptPipeline` since all the models use the same pipeline

trlx/trainer/accelerate_ilql_trainer.py
```diff
@@ -95,3 +142,73 @@ def save_pretrained(self, directory: Optional[str] = None):
             "`AccelerateILQLTrainer` does not currently support automatic saving "
             "with `transformers.PreTrainedModel.save_pretrained`."
         )
+
+    def make_experience(self, samples, rewards, max_length=2048):
```
cat-state, line 146: maybe pull this into somewhere it can be shared with nemo impl?

cat-state (changes requested):
I think this mostly looks good except for the nemo, in future we can try to refactor and break up the big make_experience functions

trlx/trlx.py
jon-tow: * Yeah I wasn't sure if it'd be too messy - let me give it a go :)

* I agree on replacing the `get_pipeline` stuff with just `PromptPipeline`; I originally did that but reverted before creating the PR to limit the scope to the issue being addressed. I can't think of any other prompt pipelines so not sure what the `_DATAPIPELINE` registry is even intended for?

reciprocated: > so not sure what the _DATAPIPELINE registry is even intended for?
It's from the time when every pipeline was ought to be a specific dataset, each registered deliberately

trlx/trainer/accelerate_ilql_trainer.py
jon-tow: I've duplicated the `make_experience` into NeMo for now. There was a subtle difference in logging whereby NeMo couldn't recognize the global rank `RANK == 0`checks forcing each rank to write tables to stdout (the fix is to just use their global rank check util).

I think it might be best to push this off to another PR because we'll need to re-visit this abstraction again for PPO. What do you think?

cat-state: Yeah, I think that makes sense for now. Maybe `torch.distributed.get_rank()` will work for both but we can revisit

cat-state (approved):
Thanks jon! I think this looks good to me, we can excise and clean up get_pipeline in future work and revisit the sharing between accelerate and nemo

Issue #204, opened 2023-01-18
Support for T5 for ILQL


loganlebanoff:
### 🐛 Describe the bug

It seems that there is not support for seq2seq models when using the ILQL algorithm.

```
Traceback (most recent call last):
  File "/home/local/AA/logan.lebanoff/.pycharm_helpers/pydev/pydevd.py", line 1496, in _exec
    pydev_imports.execfile(file, globals, locals)  # execute the script
  File "/home/local/AA/logan.lebanoff/.pycharm_helpers/pydev/_pydev_imps/_pydev_execfile.py", line 18, in execfile
    exec(compile(contents+"\n", file, 'exec'), glob, loc)
  File "/home/local/AA/logan.lebanoff/slamdunk/src/slamdunk/trl/test_trlx.py", line 311, in <module>
    main(args)
  File "/home/local/AA/logan.lebanoff/slamdunk/src/slamdunk/trl/test_trlx.py", line 289, in main
    model = trlx.train(
  File "/home/local/AA/logan.lebanoff/trlx/trlx/trlx.py", line 104, in train
    trainer = get_trainer(config.train.trainer)(
  File "/home/local/AA/logan.lebanoff/trlx/trlx/trainer/accelerate_ilql_trainer.py", line 16, in __init__
    super().__init__(config, **kwargs)
  File "/home/local/AA/logan.lebanoff/trlx/trlx/trainer/accelerate_base_trainer.py", line 51, in __init__
    self.model = self.setup_model()
  File "/home/local/AA/logan.lebanoff/trlx/trlx/trainer/accelerate_base_trainer.py", line 100, in setup_model
    model = self.get_arch(self.config)
  File "/home/local/AA/logan.lebanoff/trlx/trlx/trainer/accelerate_ilql_trainer.py", line 32, in get_arch
    return CausalLMWithValueHeads(
  File "/home/local/AA/logan.lebanoff/trlx/trlx/trainer/nn/ilql_models.py", line 221, in __init__
    self.base_model = transformers.AutoModelForCausalLM.from_pretrained(config)
  File "/home/local/AA/logan.lebanoff/miniconda3/envs/slamdunk/lib/python3.10/site-packages/transformers/models/auto/auto_factory.py", line 466, in from_pretrained
    raise ValueError(
ValueError: Unrecognized configuration class <class 'transformers.models.t5.configuration_t5.T5Config'> for this kind of AutoModel: AutoModelForCausalLM.
[...]

PR #290, opened 2023-02-08
Support ILQL for T5 model, Fix PPO T5 for refactored code

PhungVanDuy:
Adapted ILQL implementation for Causal models, this PR support ILQL for the T5 model. 
- Related issue: https://github.com/CarperAI/trlx/issues/204

- [x] Add ILQL for the T5 model
- [x] Run Summarization example by OpenAI dataset
- [x] Add sentiment examples for PPO T5
- [x] Add tests

ILQL T5 on OpenAI Dataset: https://wandb.ai/pvduy/trlx/runs/sm5gug89
Sentiment T5: https://wandb.ai/pvduy/trlx/runs/vgvph8cc


configs/ilql_summarize_t5.yml
```diff
@@ -0,0 +1,52 @@
...
+  checkpoint_interval: 100000
+  eval_interval: 500
+
+  pipeline: "PromptPipeline"
+  trainer: "AccelerateILQLTrainer"
+  checkpoint_dir: "gpt-j-ilql-summarize-duy"
+  #  tracker: null
+
+model:
+  model_path: "/admin/home-duyphung/software/github/pvd/trlx/examples/summarize_rlhf/flan-t5-qa/checkpoint-3000"
```
reciprocated, line 16: this path is unfortunately private, but you don't have to use sft and can just start from hf's checkpoint here

examples/summarize_rlhf/ilql_summarize_t5.py
```diff
@@ -0,0 +1,90 @@
...
+            )
+            input_ids = encodings_dict["input_ids"].to(rw_device)
+            attn_masks = encodings_dict["attention_mask"].to(rw_device)
+            input_ids = input_ids.repeat(2, 1)
+            attn_masks = attn_masks.repeat(2, 1)
+            with torch.no_grad():
+                sub_scores = rw_model(input_ids=input_ids, attention_mask=attn_masks)
+            scores_list.append(sub_scores["chosen_end_scores"])
+        scores = torch.cat(scores_list, dim=0)
+        return scores
```
reciprocated, line 58: maybe we can reshare code for the reward model instantiation with ppo_t5 as well

trlx/orchestrator/offline_orchestrator.py
```diff
@@ -0,0 +1,204 @@
+import os
```
reciprocated, line 1: the contents of this file were refactored back into `accelerate_ilql_trainer.py` and the file itself was deleted

trlx/trainer/nn/ilql_models.py
```diff
@@ -62,12 +63,15 @@ class ILQLConfig(MethodConfig):
...
         logits, (qs, target_qs, vs) = outputs
         terminal_mask = labels.dones[:, :-1]
         n_nonterminal = max(1, terminal_mask.sum())
-
-        actions = labels.input_ids[:, 1:].gather(dim=1, index=labels.actions_ixs).unsqueeze(-1)
+        # check type of labels
+        if isinstance(labels, ILQLBatch):
+            actions = labels.input_ids[:, 1:].gather(dim=1, index=labels.actions_ixs).unsqueeze(-1)
+        else:
+            actions = labels.decoder_input_ids[:, 1:].gather(dim=1, index=labels.actions_ixs).unsqueeze(-1)
```
reciprocated, line 74: why do we want to `gather` here, wouldn't `decoder_input_ids` be already compiled in the needed format from `make_experience`

trlx/trainer/nn/ilql_models.py
```diff
@@ -331,3 +335,144 @@ def dummy_inputs(self):
...
+        Generates samples akin to hf's `.generate` but with custom logp prepossessing:
+        changing token probabilities as to how advantageous they would be
+        according to value functions estimations.
+        """
+        if attention_mask is None:
+            attention_mask = input_ids.not_equal(pad_token_id)
+        samples = input_ids.clone()
+        max_new_tokens = min(max_new_tokens, max_length - input_ids.shape[1])
+        if decoder_input_ids is None:
+            decoder_input_ids = input_ids.new_zeros(input_ids.shape[0], 1, dtype=torch.long, device=input_ids.device)
```
reciprocated, line 436: it's either `torch.zeros` or remove `dtype` and `device` in call to `new_zeros`

trlx/trainer/nn/ilql_models.py
```diff
@@ -331,3 +335,144 @@ def dummy_inputs(self):
...
+            adv = qs - vs
+            pi_beta = F.log_softmax(logits, -1)
+            pi_top_k = topk_mask(pi_beta + beta * adv, top_k)
+            pi = F.softmax(pi_top_k / temperature, -1)
+            next_tokens = torch.multinomial(pi, num_samples=1)
+            if eos_token_id is not None:
+                next_tokens = (1 - finished) * next_tokens + finished * pad_token_id
+                finished = (next_tokens == eos_token_id).long()
+            else:
+                raise ValueError("eos_token_id must be provided")
```
reciprocated, line 463: this check is great, however it's better to do it at the start of the function and maybe not in the loop

reciprocated (commented):
Glad that it works overall, but we still have to find a way to refactor many repetitions here, at least up to the same level as it's done in PPO

trlx/trainer/nn/ilql_models.py
PhungVanDuy: yes, you are right we do no need the gather here. Actions also started with 0. Will remove it, but it not effect current result.

configs/ilql_summarize_t5.yml
PhungVanDuy: Yes I will upload T5 SFT for to hf.

trlx/trainer/accelerate_ilql_trainer.py
```diff
@@ -80,12 +99,98 @@ def prepare_learning(self):
...
+            logger.info("Logging experience string statistics")
+            columns = ["Prompt Length", "Output Length", "Sample Length"]
+            table = Table(*columns, title="Experience String Stats (mean ∈ \[min, max])", show_lines=True)
+            row = []
+            for lengths in [prompt_lengths, output_lengths, sample_lengths]:
+                row.append(f"{lengths.mean():.2f} ∈ [{min(lengths)}, {max(lengths)}]")
+            table.add_row(*row)
+            Console().print(table)
+
+        returns = (returns - returns.mean()) / (returns.std() + 1e-30)
```
jon-tow, line 169: Nit: We probably shouldn't hardcode the epsilon like this at 1e-30 and just use the machine epsilon, for example`torch.finfo(returns.dtype).eps`

trlx/trainer/accelerate_ilql_trainer.py
```diff
@@ -80,12 +99,98 @@ def prepare_learning(self):
...
+            all_actions_ixs,
+            all_dones,
+        )
+
+    def make_experience(self, samples, rewards, max_length=2048):
+        """
+        Tokenizes samples and shapes rewards into proper tensors and then inserts the resulting dataset into the trainer
+        """
+
+        if self.config.model.model_arch_type == "seq2seq":
```
jon-tow, line 190: Aside: At some point, we need to design a better way to handle these arch-type conditionals 😅

trlx/models/modeling_ilql.py
```diff
@@ -339,3 +353,147 @@ def post_init(self, state_dict):
         self.ilql_heads.load_state_dict(state_dict, strict=False)
         del state_dict
         gc.collect()
+
+
+class AutoModelForSeq2SeqLMWithILQLHeads(PreTrainedModelWrapper):
```
jon-tow, line 358: Let's add a test for this similar to https://github.com/CarperAI/trlx/blob/e72f7d1a8008c9a994e9fe465aa4a8a7a1fb3232/tests/test_models.py#L138

trlx/trainer/accelerate_ppo_trainer.py
```diff
@@ -170,7 +174,7 @@ def loss(self, batch: PPORLBatch):
             end = start + response_length
             logprobs, values_pred, mask = (
                 logprobs[:, start:end],
-                values_pred[:, start - 1 : end - 1],
+                values_pred[:, start:end],
```
jon-tow, line 177: We need to add a test for this

jon-tow (changes requested):
Hi, @PhungVanDuy, great work! I've left some feedback to look at when you get a chance 👍  👍

trlx/data/ilql_types.py
```diff
@@ -38,6 +38,30 @@ class ILQLElement:
...
+    :param attention_mask: Attention mask. Should be a long tensor.
+    :type attention_mask: torch.Tensor
+
+    :param rewards: Rewards for each token. Should be a float tensor of same size as tokens.
+    :type rewards: torch.Tensor
+    """
+
+    input_ids: TensorType["query_size"]
+    attention_mask: TensorType["query_size"]
+    decoder_input_ids: TensorType["reward_size"]
```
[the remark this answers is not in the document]
PhungVanDuy: I prefer to separate this one for easier maintenance. We can re-design this one along when we find a way to seperate "seq2seq" like you said here: https://github.com/CarperAI/trlx/pull/290#discussion_r1122524562

jon-tow (approved):
Looks good on my end! Great work :)

PhungVanDuy:
> Looks good on my end! Great work :)

Thank you so much for your comment!

Issue #295, opened 2023-02-09
devcontainer fails because pylance can't be installed


derpyplops:
### 🐛 Describe the bug

When I tried building the devcontainer with vscode on my m2 mac, I was only able to install it after uncommenting pylance from the pip installs. Then when I try to install it I get 

```
$ pip install pylance
Defaulting to user installation because normal site-packages is not writeable
ERROR: Could not find a version that satisfies the requirement pylance (from versions: 0.0.2)
ERROR: No matching distribution found for pylance
```

jon-tow:
Hi @derpyplops ! The Pylance language server isn't pip installable so it should be removed from the Dockerfile. The [`Python` extension](https://marketplace.visualstudio.com/items?itemName=ms-python.python) should automatically install Pylance for you - can you check your installed VS Code extensions to see if it's there?

jon-tow:
Thanks for reporting 🙏

PR #296, opened 2023-02-09
Remove `pylance` installs from devcontainer

jon-tow:
* Removes the `pip install pylance` command in the dev Dockerfile. This installs a dependency (https://pypi.org/project/pylance/) unrelated to Microsoft Pylance language server when it doesn't error out.
* Removes `"ms-python.vscode-pylance"` from customizations as it's already included in the installation of `"ms-python.python"` (https://marketplace.visualstudio.com/items?itemName=ms-python.python).

Related Issue: #295

CC @alan-cooney for any suggestions.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -43,7 +43,6 @@ RUN pip install \
   autopep8 \
   jedi \
   mypy \
-  pylance \
   pytest \
   toml \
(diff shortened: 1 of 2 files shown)


derpyplops:
LGTM

alan-cooney:
Makes sense!

PR #306, opened 2023-02-13
Move to Python config classes instead of `ymls`

cat-state:
This change removes the default yml configs in favor of `trlx.data.default_configs.default_{ppo_config, ilql_config, sft_config}`. I tested `{ppo, ilql, sft}_sentiements`.  

There could be further changes after this change: 
* The `name` field is superflous when specifying the config on via python so we could remove it and use it only when deserializing configs.
* The same applies to the `*Trainers` and pipeline=`PromptPipeline`

Those fields are mainly needed for our handwritten deserialization, and I think we can have them be automatically be picked from the classes, at least if we use e.g `cattrs.unstructure`/`cattrs.structure` to deserialize into our typed config.


jon-tow (changes requested):
This looks great! There are still some dangling YAML config stuff.
- Remove https://github.com/CarperAI/trlx/blob/remove_ymls/configs/test_config.yml
- Remove or update https://github.com/CarperAI/trlx/blob/remove_ymls/tests/test_configs.py which ensured YAML configs were valid
- Remove lingering YAML configs within the examples dir:
  - https://github.com/CarperAI/trlx/tree/remove_ymls/examples/randomwalks/configs
  - https://github.com/CarperAI/trlx/tree/remove_ymls/examples/experiments/grounded_program_synthesis/configs

examples/sft_sentiments.py
```diff
@@ -1,26 +1,20 @@
...
+from trlx.data.default_configs import default_sft_config
 
 
 def get_positive_score(scores):
     "Extract value associated with a positive sentiment from pipeline's output"
     return dict(map(lambda x: tuple(x.values()), scores))["POSITIVE"]
 
 
-def main(hparams={}):
-    config = TRLConfig.update(default_config, hparams)
```
reciprocated, line 23: Both hyperparamer optimization (wandb's and ray tune's) relied on this config propagation, either through passing parameters updates on cli or by importing main directly and passing it within an interpreter. Not sure is there a better way of doing it, althought I feel that something emulating hydra's `+` operator is still essential

cat-state: Ah I see, I'll add them back. Yeah, mergability is easy to add back.

cat-state:
> This looks great! There are still some dangling YAML config stuff.
> 
>     * Remove https://github.com/CarperAI/trlx/blob/remove_ymls/configs/test_config.yml
> 
>     * Remove or update https://github.com/CarperAI/trlx/blob/remove_ymls/tests/test_configs.py which ensured YAML configs were valid
> 
>     * Remove lingering YAML configs within the examples dir:
>       
>       * https://github.com/CarperAI/trlx/tree/remove_ymls/examples/randomwalks/configs
>       * https://github.com/CarperAI/trlx/tree/remove_ymls/examples/experiments/grounded_program_synthesis/configs

Thanks. I address those. For the `test_config.yml` - do we still want to preserve any serialization ability? IMO, config serialization/deserialization is still useful so we should keep that testing (and move to some library for the ser/des)

jon-tow:
> IMO, config serialization/deserialization is still useful so we should keep that testing (and move to some library for the ser/des
Yup. That's a good idea; let's keep it 👍

cat-state:
Should be mergable now.

trlx/data/configs.py
```diff
@@ -1,20 +1,22 @@
...
-from typing import Any, Dict, Optional, Set
+from typing import Any, Dict, Optional
 
 import yaml
 
 from trlx.data.method_configs import MethodConfig, get_method
 
 
-def merge(base: Dict, update: Dict, updated: Set) -> Dict:
+def merge(base: Dict, update: Dict) -> Dict:
```
jon-tow, line 10: Just a note: the sweep utils are broken because the sweep configs provide a single key and I believe this and the previous implementation check the whole key-path for merging. E.g. you can't do the following merge:  
```python
merge({'optimizer': 'kwargs': 'lr': 1e-3}, {'lr': 9}) != {'optimizer': 'kwargs': 'lr': 9}
```
https://colab.research.google.com/drive/1xEhbdJjuw40ow1ObhiWmAfD1wt2R6zb2?usp=sharing

Probably best to handle this from the sweep side with a manual `evolve`. CC @reciprocated who's probably aware of this but thought I'd tag if otherwise.

jon-tow (approved):
Looks great to me! Just left a tiny note.

trlx/data/configs.py
reciprocated: I've updated it to be of the following form: https://github.com/CarperAI/trlx/blob/deb667a00cb66a1dd20f9a61ea6d7482637ac6e3/configs/sweeps/ppo_sweep.yml#L9-L14
https://github.com/CarperAI/trlx/blob/deb667a00cb66a1dd20f9a61ea6d7482637ac6e3/trlx/data/configs.py#L275-L290

cat-state: > side with a manual `evolve`. CC @reciprocated who's probably aware of this but thought I'd tag if otherwise.

Oh, I thought was actually a bug rather than feature. I see. I suppose if this is to be changed/fixed in @reciprocated 's PR I'll make a new merge function for the current use

README.md
```diff
@@ -57,6 +57,19 @@ trainer = trlx.train('EleutherAI/gpt-j-6B', samples=['dolphins', 'geese'], rewar
...
+
+```python
+from trlx.data.default_configs import default_ppo_config, TrainConfig
+
+config = default_ppo_config()
+config.model.model_path = 'EleutherAI/gpt-neox-20b'
+config.train.seq_length = 32
+config.train.batch_size = 16
+
+trainer = trlx.train(config=config, reward_fn=lambda samples, **kwargs: [float(int(sample)) for sample in samples])
```
reciprocated, line 70: Nit: I'm sure this example is not meant to be taken literally, but I still have some suggestions here :blush:: 
- `float(int(sample))` will throw a `ValueError`, maybe let's just take the `reward_fn` from the above
- `model_path` is updated to neox-20b, but the default ppo_config still has gpt2 tokenizer
- maybe something like `batch_size=1` and `seq_length=2048` will be more consistent with 20b model

reciprocated (commented):
Nice! Although to double check: are old configs still functional? And are you planning to update HH's configs, because there are way too many of them 😂

cat-state:
> Nice! Although to double check: are old configs still functional? And are you planning to update HH's configs, because there are way too many of them joy

Yes, the old ones should still work fine and still deserialize via load_yaml or from_dict

README.md
cat-state: Ah good catch! no its a good idea to have examples copy-pastable working

congchan:
Hi, I am new to trlx. Just wondering what is the best way to set params? 
Some examples use handwritten python config class, while some make use of yml file.
More specifically, I am trying the `example/summarize_rlhf`, and found that `batch size` parameters is not under controlled, no matter how I modified the `train:batch_size`, or deepspeed configs `train_micro_batch_size_per_gpu`, or `"gradient_accumulation_steps"`, the output log are always `train_batch_size ............. 2` and `train_micro_batch_size_per_gpu  2`, `"gradient_accumulation_steps": 1,`

```
[2023-03-03 17:14:21,660] [INFO] [config.py:1013:print]   optimizer_legacy_fusion ...... False
[2023-03-03 17:14:21,660] [INFO] [config.py:1013:print]   optimizer_name ............... None
[2023-03-03 17:14:21,660] [INFO] [config.py:1013:print]   optimizer_params ............. None
[2023-03-03 17:14:21,660] [INFO] [config.py:1013:print]   pipeline ..................... {'stages': 'auto', 'partition': 'best', 'seed_layers': False, 'activation_checkpoint_interval': 0}
[2023-03-03 17:14:21,660] [INFO] [config.py:1013:print]   pld_enabled .................. False
[2023-03-03 17:14:21,660] [INFO] [config.py:1013:print]   pld_params ................... False
[2023-03-03 17:14:21,660] [INFO] [config.py:1013:print]   prescale_gradients ........... False
[2023-03-03 17:14:21,660] [INFO] [config.py:1013:print]   scheduler_name ............... None
[2023-03-03 17:14:21,660] [INFO] [config.py:1013:print]   scheduler_params ............. None
[2023-03-03 17:14:21,660] [INFO] [config.py:1013:print]   sparse_attention ............. None
[2023-03-03 17:14:21,660] [INFO] [config.py:1013:print]   sparse_gradients_enabled ..... False
[2023-03-03 17:14:21,660] [INFO] [config.py:1013:print]   steps_per_print .............. inf
[2023-03-03 17:14:21,660] [INFO] [config.py:1013:print]   train_batch_size ............. 2
[2023-03-03 17:14:21,660] [INFO] [config.py:1013:print]   train_micro_batch_size_per_gpu  2
[2023-03-03 17:14:21,660] [INFO] [config.py:1013:print]   use_node_local_storage ....... False
[2023-03-03 17:14:21,660] [INFO] [config.py:1013:print]   wall_clock_breakdown ......... False
[2023-03-03 17:14:21,660] [INFO] [config.py:1013:print]   world_size ................... 1
[2023-03-03 17:14:21,660] [INFO] [config.py:1013:print]   zero_allow_untested_optimizer  True
[2023-03-03 17:14:21,660] [INFO] [config.py:1013:print]   zero_config .................. stage=2 contiguous_gradients=True reduce_scatter=True reduce_bucket_size=500,000,000 allgather_partitions=True allgather_bucket_size=500000000 overlap_comm=False load_from_fp32_weights=True elastic_checkpoint=False offload_param=DeepSpeedZeroOffloadParamConfig(device='cpu', nvme_path=None, buffer_count=5, buffer_size=100,000,000, max_in_cpu=1,000,000,000, pin_memory=False) offload_optimizer=DeepSpeedZeroOffloadOptimizerConfig(device='cpu', nvme_path=None, buffer_count=4, pin_memory=False, pipeline=False, pipeline_read=False, pipeline_write=False, fast_init=False) sub_group_size=1,000,000,000 cpu_offload_param=None cpu_offload_use_pin_memory=None cpu_offload=None prefetch_bucket_size=50,000,000 param_persistence_threshold=100,000 model_persistence_threshold=sys.maxsize max_live_parameters=1,000,000,000 max_reuse_distance=1,000,000,000 gather_16bit_weights_on_model_save=False stage3_gather_fp16_weights_on_model_save=False ignore_unused_parameters=True legacy_stage1=False round_robin_gradients=False
[2023-03-03 17:14:21,660] [INFO] [config.py:1013:print]   zero_enabled ................. True
[2023-03-03 17:14:21,660] [INFO] [config.py:1013:print]   zero_optimization_stage ...... 2
[2023-03-03 17:14:21,660] [INFO] [config.py:998:print_user_config]   json = {
    "train_micro_batch_size_per_gpu": 2,
    "gradient_accumulation_steps": 1,
    "fp16": {
        "enabled": true,
        "min_loss_scale": 0.5,
        "fp16_scale_tolerance": 0.25,
        "opt_level": "O2",
        "auto_cast": false
    },
    "zero_optimization": {
        "stage": 2,
        "offload_param": {
            "device": "cpu"
        },
        "offload_optimizer": {
            "device": "cpu"
        },
        "allgather_partitions": true,
        "allgather_bucket_size": 5.000000e+08,
        "contiguous_gradients": true
    },
    "steps_per_print": inf,
    "bf16": {
        "enabled": false
    },
    "zero_allow_untested_optimizer": true
}
```

Issue #223, opened 2023-01-25
Remove yaml config files


alan-cooney:
### 🚀 The feature, motivation, and pitch

Was there a reason for using them, or just a legacy feature of TRL?

Advantages of using the Python data class directly are:
- Type safety (it's very easy to make typos in yaml, that Python can catch)
- One less language (most people know yaml to some degree but not well, some won't have used it)
- Code hints (e.g. in VS Code) -> better dev experience

This can be non-breaking (i.e. just change it for the defaults & also the examples, but let users keep parsing their yml files if they already exist). Also seems like a good first issue for someone who wants to start contributing.

LouisCastricato:
We've discussed moving to the hydra config system for a while, I think it's on the back burner right now though

alan-cooney:
Thanks for the context!

Hydra seems like overkill, so I think that makes sense to not implement it. It's also a pretty old-school approach (i.e. generally the direction of travel is YML -> type safe config as code, not the other way round). 

I think this would instead be a significant improvement to the codebase, but of course up to the maintainers.

cat-state:
I agree with this wholeheartedly.

Mistobaan:
+1 on the no-config files 
-1 on hydra seems overkill for trlx.

PR #312, opened 2023-02-16
[feat] Ray train integration

reciprocated:
This PR lets Ray Tune use Ray's `AccelerateTrainer` from @yard1


ayulockin:
Hey @reciprocated, do you need any help with this PR?

trlx/ray_train/accelerate_trainer.py
```diff
@@ -169,7 +169,7 @@ def wrapped_train_loop_per_worker(*args, **kwargs):
                 os.environ["LOCAL_RANK"] = str(session.get_local_rank())
                 os.environ["LOCAL_WORLD_SIZE"] = str(session.get_local_world_size())
                 os.environ["LOCAL_SIZE"] = str(session.get_local_world_size())
-                os.environ["ACCELERATE_TORCH_DEVICE"] = str(get_device())
+                os.environ["ACCELERATE_TORCH_DEVICE"] = f"cuda:{session.get_local_rank()}"
```
Yard1, line 172: qq, what's the reason for this change?

reciprocated: this is a crude fix for device mismatch errors which happen otherwise

Yard1: Are you manually setting the devices anywhere else?

reciprocated:
@ayulockin Nothing comes to mind right now, but I'll be sure to ping you when we'd need something 😊

trlx/sweep.py
```diff
@@ -89,9 +323,39 @@ def tune_function(train_function, param_space: dict, tune_config: dict, resource
...
+    tuner = tune.Tuner(
+        AccelerateTrainer(
+            script.main,
+            # Mandatory arg. None means use Accelerate default path
+            accelerate_config_path=args.accelerate_config,
+            scaling_config=ScalingConfig(
+                trainer_resources={"CPU": 0},
+                num_workers=args.num_gpus,
+                use_gpu=True,
+                resources_per_worker={"CPU": args.num_cpus},
```
Yard1, line 343: just to be a bit more explicit
```suggestion
                resources_per_worker={"CPU": args.num_cpus, "GPU": 1},
```

trlx/ray_train/accelerate_trainer.py
reciprocated: Interestingly we do not, however with `get_device` there are always some discrepancies
as for example with deepspeed: 
```
os.environ.get("LOCAL_RANK", 0)='4' 
self.accelerator.device=device(type='cuda', index=3)
next(self.model.parameters()).device=device(type='cuda', index=4)
```

and even with ddp: 
```
os.environ.get("LOCAL_RANK", 0)='1'
batch.input_ids.device=device(type='cuda', index=7)
self.accelerator.device=device(type='cuda', index=7)
next(self.model.parameters()).device=device(type='cuda', index=7)
```
there is still an error even though batch is deliberately transfered https://github.com/CarperAI/trlx/blob/e72f7d1a8008c9a994e9fe465aa4a8a7a1fb3232/trlx/trainer/accelerate_ilql_trainer.py#L57 

```
File "/trlx/examples/ilql_sentiments.py", line 36, in main
  trlx.train(
File "/trlx/trlx/trlx.py", line 119, in train
  trainer.learn()
File "/trlx/trlx/trainer/accelerate_base_trainer.py", line 477, in learn
  loss, stats = self.loss(batch)
File "/trlx/trlx/trainer/accelerate_ilql_trainer.py", line 55, in loss
  logits, qs, target_qs, vs, _ = self.model(
File "/trlx/.env/lib/python3.9/site-packages/torch/nn/modules/module.py", line 1194, in _call_impl
  return forward_call(*input, **kwargs)
File "/trlx/.env/lib/python3.9/site-packages/accelerate/utils/operations.py", line 489, in __call__
  return convert_to_fp32(self.model_forward(*args, **kwargs))
File "/trlx/.env/lib/python3.9/site-packages/torch/amp/autocast_mode.py", line 14, in decorate_autocast
  return func(*args, **kwargs)
File "/trlx/.env/lib/python3.9/site-packages/torch/nn/parallel/distributed.py", line 1040, in forward
  output = self._run_ddp_forward(*inputs, **kwargs)
File "/trlx/.env/lib/python3.9/site-packages/torch/nn/parallel/distributed.py", line 1000, in _run_ddp_forward
  return module_to_run(*inputs[0], **kwargs[0])
File "/trlx/.env/lib/python3.9/site-packages/torch/nn/modules/module.py", line 1194, in _call_impl
  return forward_call(*input, **kwargs)
File "/trlx/trlx/trainer/nn/ilql_models.py", line 249, in forward
  out = self.base_model.transformer(**forward_kwargs)
File "/trlx/.env/lib/python3.9/site-packages/torch/nn/modules/module.py", line 1194, in _call_impl
  return forward_call(*input, **kwargs)
File "/trlx/.env/lib/python3.9/site-packages/transformers/models/gpt2/modeling_gpt2.py", line 830, in forward
  inputs_embeds = self.wte(input_ids)
File "/trlx/.env/lib/python3.9/site-packages/torch/nn/modules/module.py", line 1194, in _call_impl
  return forward_call(*input, **kwargs)
File "/trlx/.env/lib/python3.9/site-packages/torch/nn/modules/sparse.py", line 160, in forward
  return F.embedding(
File "/trlx/.env/lib/python3.9/site-packages/torch/nn/functional.py", line 2210, in embedding
  return torch.embedding(weight, input, padding_idx, scale_grad_by_freq, sparse)
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:7 and cuda:1! (when checking argument for argume
nt index in method wrapper__index_select)
```

Yard1: I see, thanks. How can I reproduce this?

trlx/sweep.py
```diff
@@ -1,83 +1,317 @@
...
         type=str,
         required=True,
         help="The config file defining the param_space.",
     )
-    parser.add_argument("--num-cpus", type=int, default=4, help="Number of CPUs to use per exp.")
-    parser.add_argument("--num-gpus", type=int, default=1, help="Number of GPUs to use per exp.")
-    parser.add_argument("-y", "--assume-yes", action="store_true", help="Don't ask for confirmation")
     parser.add_argument(
-        "--server-address",
+        "--default_config",
```
jon-tow, line 279: Do we plan on removing this `default_config`  arg option now that the PR for removing config YAMLs has been merged (#223)? If this is a blocker, I wouldn't be completely opposed to maintaining the leftover configs for now since things are backward-compatible. In a later release, we can fully deprecate config file usage.

jon-tow (commented):
Amazing work @reciprocated and @Yard1! 

I was able to run the single node (4 GPUs) example but not yet the multi-node sweep due to the lack of compute availability on our cluster. Looks great 👍  I've left a comment on the potential merge conflicts we'll need to resolve.

trlx/ray_train/accelerate_trainer.py
jon-tow: @reciprocated I'm not familiar with the Ray session API but I'm curious if this might have an effect on multi-node sweeps - have you noticed any changes there? (probably shouldn't but, again, just curious 😄)

Yard1: It has been designed to work seamlessly with multi-node setups! There should be no issues there.

One thing that is concerning is that device situation, though. Ideally we would delegate picking the device to Ray Train (which is what setting that env var achieved). I'll have to see how I can reproduce this to get to the bottom of the mismatch issue.

reciprocated: @Yard1 you can reproduce with:
```
git checkout test-device-mismatch
python -m trlx.sweep -y --config configs/sweeps/ppo_sweep.yml --default_config configs/ppo_config.yml --accelerate_config configs/accelerate/zero2-bf16.yaml --num_gpus 4 examples/ppo_sentiments.py
```

@jon-tow no, I hadn't seen anything abnormal when running multi-node

configs/sft_config.yml
```diff
@@ -29,8 +29,8 @@ optimizer:
 scheduler:
   name: "cosine_annealing"
   kwargs:
-    T_max: 10000 # train.total_steps
-    eta_min: 1.0e-4
+    T_max: 100000000000
```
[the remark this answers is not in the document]
reciprocated: Yeah.. this is scheduler: "disabled"

trlx/sweep.py
reciprocated: it's a dilemma whether to convert the rest of configs, or support two ways of updating the "default" one within sweeps. Although I'm already midway with the latter (on the examples which hadn't been converted), it's actually probably more sane to do the former instead.

jon-tow: If you're already ahead on the conversions then that's fine too. Let me know if I can help out with them. Which ones still need updating?

Yard1:
FYI we'll be able to remove a good chunk of the code from `launch.py` after https://github.com/huggingface/accelerate/pull/1123 is merged into Accelerate

trlx/sweep.py
reciprocated: that would be summarization's, but I think it would make sense to update configs from the updated branch and also after the most recent fix

Yard1:
I'd like to look into that device mismatch issue and update the `launch.py` to take advantage of the PR that was merged into Accelerate. I'll have some time tomorrow or on Friday. Can we hold off on merging before that?

reciprocated:
Yeah absolutely, sounds good!

Yard1:
Hey, sorry it took so long. I believe I fixed the issue with the device mismatch. I'll make a new PR into this branch.

Also, we may want to consider requiring accelerate>=0.17.0 for this feature (not yet released, has to be installed from master ATM). That release will include https://github.com/huggingface/accelerate/pull/1123, which will allow for a lot of code to be removed. I will include that change in my PR as well. Let me know what you think.

Yard1:
FYI we are working on integrating the trainer directly into Ray here - https://github.com/ray-project/ray/pull/33269

Yard1:
Hi, we have merged https://github.com/ray-project/ray/pull/33269.

This PR can be updated to remove `AccelerateTrainer` from TRLX and instead use `ray.train.huggingface.accelerate.AccelerateTrainer`, after pinning to (tomorrow's) Ray nightly wheel.

trlx/sweep.py
```diff
@@ -1,83 +1,310 @@
...
     tune_config = get_tune_config(config.pop("tune_config"))
     param_space = get_param_space(config)
+    column_names = list(param_space.keys())
+    target_metric = tune_config["metric"]
 
-    # Initialize Ray.
     if args.server_address:
         ray.init(address=f"ray://{args.server_address}")
     else:
         ray.init()
```
jon-tow, line 306: Running the example script from the `README.md` within a singe node 8xA100 instance (CoreWeave) I get `ConnectionError`s of the kind:
```sh
ConnectionError: Could not read 'session_name' from GCS. Did GCS start successfully?
```
(obviously not using GCS)

Passing the `local` flag to this init (`ray.init("local")`) seems to address the issue. So, I'm wondering if maybe we should include a CLI arg for such use cases or simply update the README to tell users to start a ray instance first and then run the command. Has any PR author run into the `ConnectionError` on the most recent commit?

Note: This is in a fresh env from this branch.

jon-tow (commented):
Great to finally be able to scale these sweeps! Collapsing the `ray_tune` module into `sweep.py` is very clean 🤗

I left a small concern on the README example if y'all could take a look when possible.

trlx/sweep.py
Yard1: `ray.init()` just by itself should work fine, and it will automatically start a Ray cluster composed of just the current instance if one is not running already. Are you setting the `server_address` arg?

jon-tow: I no longer have this issue. There have been some network issues with our cluster recently so I'm going to chalk it up to that. Note that I did not set the `server_address` arg for what it's worth. 
Re GCS:  I quickly assumed it was Google Cloud related, thank you for clarifying 🙏

reciprocated: > I no longer have this issue
@jon-tow How was it resolved for you? I experience the same thing right now, never encountering it before. Something must messing with ray autodiscovery, either starting a ray cluster by hand or forcing `local` does work however.

jon-tow: I'm not quite sure. I tried a new node and it was then able to auto-launch locally 😅

jon-tow (approved):
Looks good on my end! (@reciprocated cleared up an issue I ran into with disk syncing so my concerns have been addressed outside here).

ayulockin:
Hey @reciprocated now that this PR is merged we will have to update this W&B report: https://wandb.ai/ayut/trlx-ppo-sentiments-hyperopt/reports/RLHF-Hyperparameter-Optimization-for-trlX--VmlldzoyOTgxMTQ2

Do we have a tutorial or example I can refer for the changes. :)

reciprocated:
@ayulockin Sure thing! However I don't think anything has changed functionality-wise, except for this remark from the report:
> Currently distributed training is not supported and for that reason using only one GPU per trial is recommended

which has now become outdated :)

PR #128, opened 2022-12-08
Add helpful and harmless example

reciprocated:
This PR adds an example of training on Anthropic's [dataset](https://github.com/anthropics/hh-rlhf)

PPO: https://wandb.ai/sorry/trlx/runs/ab0ehsxm
SFT: https://wandb.ai/sorry/trlx/runs/0aulb18k
ILQL: https://wandb.ai/sorry/trlx/runs/30ab7epk

PR #322, opened 2023-02-21
[feat] Add `batch_size` option for the reward model

maxreciprocate:
This PR add's new config option to control `batch_size` used in `reward_fn`. Needed to make custom configs for different setups with different amount of memory available (80GB, 40GB or even cpu offload).

https://api.wandb.ai/links/sorry/g03gp4g7

### DIFF
--- a/configs/ilql_config.yml
+++ b/configs/ilql_config.yml
@@ -3,6 +3,7 @@ train:
   batch_size: 128
   epochs: 100
   total_steps: 1000
+  reward_model_batch_size: 256
 
   checkpoint_interval: 1000
   eval_interval: 100
--- a/configs/ppo_config.yml
+++ b/configs/ppo_config.yml
@@ -3,6 +3,7 @@ train:
   epochs: 100
   total_steps: 10000
   batch_size: 128
+  reward_model_batch_size: 256
 
   checkpoint_interval: 10000
   eval_interval: 100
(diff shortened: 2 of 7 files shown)


cat-state:
Since this is a configuration item only relevant to the examples and not to the trainers themselves, I'm not sure if it should actually be trlx's concern what batchsize you use for your RM

reciprocated:
I agree this configuration item is not of necessity for trainer (unless we make it so by batching inside the trainer, unlike as currently outside), but it's rather for convenience since it's a reoccurring item, perhaps not deserving to be put into ad-hoc category of using environment variables or argvs for it. Yet I'm not married to this, it's just something that came up with many configs of https://github.com/CarperAI/trlx/pull/128/files

LouisCastricato:
Can we merge or close this

Dahoas:
Same question as @LouisCastricato. I think it makes sense to push batching functionality onto the the `reward_fn` designer, as is done in the [HH example](https://github.com/CarperAI/trlx/blob/404217b2f3f295ff0f68851524517064acc43a15/examples/hh/ppo_hh.py#L170).

Dahoas:
@maxreciprocate are you ok with closing this?

maxreciprocate:
Yeah, I'm closing it

PR #331, opened 2023-02-23
Remove example usage of deprecating `trlx.train` dataset arg

jon-tow:
Quick fix to remove all `dataset` args to `trlx.train` to avoid promoting deprecating usage.

- If `dataset` is provided, the default behavior is to load the SFT config even though ILQL could be the intended method (see simulacra notebook). 
https://github.com/CarperAI/trlx/blob/165422d1a90eeef2a4a87647477ba5e8703af79b/trlx/trlx.py#L57-L62


cat-state (approved):
Good catch!

PR #333, opened 2023-02-23
Use attrs/cattrs instead of custom serialization

cat-state:
Depends on https://github.com/CarperAI/trlx/pull/306
This PR adds cattrs and attrs usage to remove us having to write custom ser/des code.

To keep supporting the ability to define `MethodConfig`s and `*Trainer`s outside TRLX however, we need to keep some of the custom serialization/deserialization (rather than just making a `Union` of all the relevant config types and having cattrs deserialize it)


cat-state:
Converting to draft for discussion around refactoring and config breakage.

PR #346, opened 2023-03-01
[fix] Convert the rest of configs from `ymls`

reciprocated:
This PR converts randomwalks and hh examples' configs. This change is a prerequisite for sweep integration and also benchmark testing

- [x] convert configs
- [x] remove yml configs

https://wandb.ai/sorry/trlx/reports/Convert-the-rest-of-configs-from-ymls-346--VmlldzozNjc4NjA4


examples/hh/ilql_hh.py
```diff
@@ -1,11 +1,71 @@
 import os
 
-import yaml
 from datasets import load_dataset
 from ppo_hh import create_reward_fn
 
 import trlx
-from trlx.data.configs import TRLConfig
+from trlx.data.default_configs import (
```
cat-state, line 7: Now I see this pattern i wonder if we should just merge all the non-method configs into just TRLConfig since they're always used together

cat-state (approved):
LGTM. I think it should be fine to delete the most of the yaml configs but we should probably keep one around as an example/testing? Since we still want to support that use case for @LouisCastricato

LouisCastricato:
Yeah keep some around for testing

reciprocated:
There is https://github.com/CarperAI/trlx/blob/main/configs/test_config.yml

PR #350, opened 2023-03-03
fix default_ilql_config in notebook

xu-song:


jon-tow (approved):
Thank you for the fix @xu-song!

Issue #282, opened 2023-02-06
Add NeMo trainer for SFT


jon-tow:
### 🚀 The feature, motivation, and pitch

Following #280, add support for NeMo-based supervised fine-tuning. 

NeMo Working Group: https://discord.com/channels/981279233835958313/1072219122827530302

PR #353, opened 2023-03-03
Add Supervised Fine-Tuning (SFT) support for NeMo backend

jon-tow:
**Report**: https://wandb.ai/carperai/trlx-nemo-sft/runs/uzzbp96e
**Related Issue**: #282


configs/nemo_configs/sft_megatron_20b.yaml
```diff
@@ -0,0 +1,148 @@
+name: megatron_gpt
+restore_from_path: null  # used when starting from a .nemo file
+
+trainer:
+  devices: 8
+  num_nodes: 4
+  accelerator: gpu
+  precision: bf16
```
cat-state, line 8: did u notice any issue w bf16 vs fp16?

trlx/models/modeling_nemo_sft.py
```diff
@@ -0,0 +1,523 @@
...
+            output_tensor = model(
+                input_ids=input_ids,
+                position_ids=position_ids,
+                attention_mask=attention_mask,
+                checkpoint_activations_all_layers=checkpoint_activations_all_layers,
+            )
+
+            def loss_func(output_tensor):
+                # Shift logits and labels to align predictions
+                logits = output_tensor[:, :-1, :]
```
cat-state, line 435: When using SP, are the outputs already gathered somewhere within the model? Since otherwise the tokens at the end of the intermediate SP chunks will get lost (hence the gather_from_sequence_parallel_region in ILQL)

trlx/models/modeling_nemo_sft.py
```diff
@@ -0,0 +1,523 @@
...
+    def parameters(self):
+        return (p for p in self.model.parameters() if p.requires_grad)
+
+    def build_attention_mask_and_position_ids(
+        self, data: torch.LongTensor, reset_attention_mask: bool = False
+    ) -> Tuple[torch.BoolTensor, torch.LongTensor]:
+        micro_batch_size, seq_length = data.size()
+
+        # Attention mask (lower triangular).
+        if reset_attention_mask:
```
cat-state, line 399: Is this not always False?

cat-state (approved):
Thanks jon! I think this mostly LGTM aside from some questions. I also noticed in the wandb that the global step seems to drop back down to a low count every validation, which was something that I saw with ILQL but fixed and don't see anymore, so if that's easy to fix then would be good to add that!

configs/nemo_configs/sft_megatron_20b.yaml
```diff
@@ -0,0 +1,148 @@
...
+exp_manager:
+  # set this to save checkpoints
+  explicit_log_dir: sft_sentiments_logs
+  exp_dir: null
+  name: megatron_gpt_20b_sft_sentiments
+  create_tensorboard_logger: False
+  create_wandb_logger: True
+  wandb_logger_kwargs:
+    project: trlx-nemo-sft
+    name: megatron_gpt_20b_ppo_sentiments
```
cat-state, line 32: This should be megatron_gpt_20b_sft_sentiments

configs/nemo_configs/sft_megatron_20b.yaml
jon-tow: From a few trial runs I noticed that fp16 was a bit more unstable (very large grad norms at the start - see [trlx: BF16 v FP16 - Add Supervised Fine-Tuning \(SFT\) support for NeMo backend \#353](https://api.wandb.ai/links/carperai/xyfpdy3r))  so I opted to default to bf16.

trlx/models/modeling_nemo_sft.py
```diff
@@ -0,0 +1,523 @@
...
+                    loss = tensor_parallel.vocab_parallel_cross_entropy(logits.float(), labels)
+
+                _loss_mask = _loss_mask.contiguous().view(-1).float()
+                loss = loss.transpose(0, 1).contiguous().view(-1).float()
+                loss_for_mb = torch.sum(loss * _loss_mask) / _loss_mask.sum()
+
+                reduced_loss = average_losses_across_data_parallel_group([loss_for_mb])
+
+                # TODO: figure out why this sync is needed (crashes otherwise)
+                torch.cuda.synchronize()
```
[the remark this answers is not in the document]
jon-tow: Apparently so lol I still get NCCL errors when removing it.

configs/nemo_configs/sft_megatron_20b.yaml
cat-state: oh okay thanks for checking :)

jon-tow:
Thanks for the review, Aman!

PR #358, opened 2023-03-07
Enable infinite PromptPipeline

alexandremuzio:
Add `infinite` flag to `PromptPipeline` to enable cyclic pipeline behavior.
This fixes a TODO in the `accelerate_ppo_trainer.py`.


alexandremuzio:
Seems like there's possibly a bug in the `accelerate/data_loader/BatchSamplerShard` which doesn't play well with infinite sized pipelines. Reverting to separate function `infinite_dataloader` solution

cat-state (approved):
Thanks @alexandremuzio !

PR #361, opened 2023-03-08
[fix] Update `AdaptiveKLController` with correct KL

reciprocated:
This PR corrects updates for `AdaptiveKLController`, and adds logging of a square root of an estimate of `D_KL(pi || pi_ref)`

https://wandb.ai/sorry/trlx-references/reports/fix-kl-controller-v-main--VmlldzozNzM2Mzg0

PPO runs start to diverge against references precisely at the point when KL exceeds the controller's target and its scalar multiplier increases, whereas previously the same target value would not be reached.


Dahoas (approved):
Looks good!

PR #362, opened 2023-03-10
[fix] Drop `<eos>` from ILQL sample's phrases

reciprocated:
This PR removes `<eos>` token from each sample's prompts & continuations and instead forces the last continuation to end with one, even in case of right truncation. 

https://wandb.ai/sorry/trlx-references/reports/fix-ilql-sample-endings-v-main--VmlldzozNzUyNzAw#ilql_randomwalks/gpt2config/1gpu


cat-state (approved):
LGTM

PR #363, opened 2023-03-13
fixes half exp not implemented error

Dahoas:
Fixes error `cuda-exp-vml not implemented for half on cpu`

PPO Sentiments: https://wandb.ai/dahoas/trlx/runs/ycl8x4h6/overview?workspace=user-dahoas


arpitg1991:
Thanks for this

PR #364, opened 2023-03-13
Minibatch impl

Dahoas:
Implements minibatching for PPO. 

PPO sentiments `bs: 32, mbs: 16`. https://wandb.ai/dahoas/trlx/runs/oo6t8rla/overview?workspace=user-dahoas

PPO HH on GPT-NeoX `bs: 4, mbs: 1`. https://wandb.ai/dahoas/trlx/runs/9rnbmtu6?workspace=user-dahoas


trlx/data/configs.py
```diff
@@ -196,6 +196,9 @@ class TrainConfig:
 
     :param seed: Random seed
     :type seed: int
+
+    :param minibatch_size: Size of model input during one forward pass. Must divide batch size
```
cat-state, line 200: (very pedantic) nit, feel free to ignore: usually I've heard this called _micro_ batch, with minibatch referring to what we usually call "a batch" (to distinguish from a single batch of the whole dataset)

trlx/trainer/accelerate_base_trainer.py
```diff
@@ -474,12 +480,23 @@ def learn(self):  # noqa: C901
...
-                    forward_time = time() - forward_time
-                    backward_time = time()
-                    self.accelerator.backward(loss)
-                    backward_time = time() - backward_time
+                    forward_time = 0
+                    backward_time = 0
+                    stats_accum = []
+                    for mbi in range(self.num_mb):
+                        forward_time -= time()
+                        loss, stats = self.loss(batch)
```
cat-state, line 488: Shouldn't this get loss using a minibatch sliced from `batch`?

Dahoas: Ah yeah good catch (I copied this over from a different branch and forgot to change this)

trlx/trainer/accelerate_base_trainer.py
```diff
@@ -468,18 +475,40 @@ def learn(self):  # noqa: C901
...
-                    forward_time = time()
-                    loss, stats = self.loss(batch)
-                    forward_time = time() - forward_time
-                    backward_time = time()
-                    self.accelerator.backward(loss)
-                    backward_time = time() - backward_time
+                    forward_time = 0
+                    backward_time = 0
+                    stats_accum = []
+                    for mb in mbs:
```
eluzhnica, line 497: to avoid unnecessary gradient synchronization when doing using gradient accumulation you can simply add: `self.accelerator.accumulate(self.model)` here

eluzhnica, line 497: to avoid unnecessary gradient synchronization when doing using gradient accumulation you can simply add: `self.accelerator.accumulate(self.model)` here

cat-state: fwiw i tried using that atop this PR and got weird results: https://wandb.ai/uwu1/trlx/reports/Untitled-Report--VmlldzozOTAyNDg4

cat-state (approved):
LGTM!

trlx/trainer/accelerate_base_trainer.py
eluzhnica: yes, but we're specifying that already in the config files like zero2-bf16.yaml

did you add it right below this line or literally at the top of the script? it should be added below this line since the entering of the context is how `accumulate` tracks the number of steps that are being executed so that it knows when to sync (has an internal step counter)

Also the division of the loss by self.num_mb should go away since that would be handled by accelerator

Here's an example: https://github.com/muellerzr/timing_experiments/blob/main/good.py#L152

eluzhnica: @cat-state Okay just made a [PR](https://github.com/CarperAI/trlx/pull/396). I'll try to run the same tests that @Dahoas ran to confirm it works as intended.

As for the TRLConfig, happy to do so, but it seems to me that the configs for accelerate are being set up separately in the repo (configs/accelerate/....) and that is the general pattern I've seen before too. And accelerate feeds those params behind the scenes for us automatically, so if we were to also specify in the TRLConfig it is a bit redundant (and potentially conflictual values). let me know what you think

trlx/trainer/accelerate_base_trainer.py
```diff
@@ -468,18 +475,40 @@ def learn(self):  # noqa: C901
         for _ in range(self.config.train.epochs):
             # For each batch
             for batch in self.train_dataloader:
+                mbs = [
+                    PPORLBatch(
```
eluzhnica, line 479: @Dahoas Just to confirm, this is still a draft correct? Since (although I haven't run it) this I think would break for ILQLBatch and other types of datatypes

cat-state: Ah you're right, I just ran this with the benchmarking scripts and it crashed here with this error

cat-state (changes requested):
Need to make it not incompatible with ILQL

trlx/trainer/accelerate_base_trainer.py
eluzhnica: Fixed this here: https://github.com/CarperAI/trlx/pull/403

Dahoas:
Merge #403 which makes mini-batching compatible with all trainers.

Dahoas:
Let's merge this into main once #396 gets merged.

Dahoas:
@cat-state Can you take another look and if it looks good we can merge?

cat-state:
thanks @eluzhnica and @Dahoas ! just tried w ilql and it seems to work https://wandb.ai/uwu1/trlx/runs/m2e4rwga?workspace=user-uwu1
Thanks for working together and adding this!

cat-state (approved):

PR #370, opened 2023-03-16
Add `hypothesis` tests for ILQL

cat-state:
Adds more tests for ILQL heads and data preprocessing. Also adds usage of [`hypothesis`](https://hypothesis.readthedocs.io/en/latest/) for generating test cases.

### DIFF
--- a/trlx/pipeline/offline_pipeline.py
+++ b/trlx/pipeline/offline_pipeline.py
@@ -1,4 +1,5 @@
-from typing import Iterable, List, Union
+from dataclasses import dataclass
+from typing import Iterable, List, Tuple, Union
 
 import torch
 from torch.nn.utils.rnn import pad_sequence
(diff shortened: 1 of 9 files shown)


trlx/pipeline/offline_pipeline.py
```diff
@@ -14,45 +19,53 @@
...
         dialogue = [tokenizer.bos_token, dialogue]
     elif isinstance(dialogue, tuple):
+        if len(dialogue) % 2 != 0:
+            raise ValueError("Dialogue must have an even number of phrases, alternating prompt and output")
         dialogue = list(dialogue)
 
-    out = []
-    ctx_length = max_length - 1
+    if not dialogue[-1].endswith(tokenizer.eos_token):
+        dialogue[-1] = dialogue[-1] + tokenizer.eos_token
```
reciprocated, line 42: If samples are truncated down the line, `<eos>` will be trimmed as well, so that's not the same behavior as on the main. Duy, in his experiments, said that this is required for T5

reciprocated (commented):
Thanks Aman, this is so cool actually! I've checked that tests pass and there seem to be no regressions, except for a small difference with main in trimming `<eos>`

https://wandb.ai/sorry/trlx-references/reports/add-hypothesis-v-main--VmlldzozODY0OTY2#ilql_sentiments/gpt2/1gpu

trlx/pipeline/offline_pipeline.py
cat-state: https://wandb.ai/uwu1/trlx/reports/t5-truncation-change--VmlldzozOTAwNzYw
It does seem to change it but only for the later training iterations

reciprocated (approved):
LGTM!

Issue #368, opened 2023-03-15
The tqdm bar is not displayed correctly


mcao516:
### 🐛 Describe the bug

When `self.total_steps > train.epochs * len(train_dataloader)` and the model is trained on multiple GPUs, the tqdm bar is not displayed correctly.

For example, when trained on 2 GPUs, the tqdm bar is only 50% when the training is done.

PR #374, opened 2023-03-16
[fix] ILQL `total_steps` calculation when running distributed

reciprocated:
Fix of #368, previously `total_steps` were calculated against an unsharded dataloader, wrongly giving n-times bigger estimate of the number of gradient steps in the `tqdm` bar


cat-state (approved):
LGTM!

Issue #371, opened 2023-03-16
Question about the shape of `logprobs` in the PPO algorithm


wiio12:
Hi, I have a little question about the `logprobs` variable in `PPORLElement`, located below:
https://github.com/CarperAI/trlx/blob/46231059f5a3fa256a25ec2a16f36b24b2b34055/trlx/data/ppo_types.py#L35

When I ran the code, I discovered that the `logprobs` variable is in the shape of `TensorType["response_size"]`, without the "vocab size". But no error is raised with these mismatched types.

Which one of these is correct?

wiio12:
Hi, @jon-tow, thanks for the reply! :)

PR #377, opened 2023-03-17
fix(docs): Update incorrect `PPORLElement` logprob tensor shape hint

jon-tow:
Logits are collapsed along vocab dimension for logprobs resulting in tensor shape `[..., response_size]`

https://github.com/CarperAI/trlx/blob/2f90ba0ecd640ae18cd62adb5e934a4b779f534b/trlx/trainer/accelerate_ppo_trainer.py#L433

Related Issue: #371


reciprocated (approved):
Thanks Jon!

PR #381, opened 2023-03-20
[feat] Add optional dependency list

reciprocated:
This PR adds an optional dependency list (torch==2.0.0, accelerate==0.17.1, transformers==4.27.1, etc) for reproducable runs.


cat-state (approved):
Would it be possible to make this a target in setup.cfg? then its all in one place. LGTM otherwise

reciprocated:
Reached agreement to keep pinned dependencies in requirements.txt 

[0] https://caremad.io/posts/2013/07/setup-vs-requirement/

PR #386, opened 2023-03-23
Use `AccelerateTrainer` from Ray

Yard1:
Use `AccelerateTrainer` added in Ray nightly instead of having the `AccelerateTrainer` in trlx.


Yard1:
cc @reciprocated

PR #387, opened 2023-03-23
[fix] Fix ILQL head sync under ZeRO3

reciprocated:
This PR fixes a bug where ILQL target heads were never being updated on the most recent accelerate & deepspeed versions. Removal of the chain iterator is not just a refactor, it fixed updates on heads from being only temporary and not persistent

https://wandb.ai/sorry/trlx/reports/-fix-Fix-ILQL-head-sync-under-ZeRO3-387---VmlldzozODc1MjI2

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -11,7 +11,7 @@ license = MIT
 [options]
 packages = find:
 install_requires =
-    accelerate>=0.12.0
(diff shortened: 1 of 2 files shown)


cat-state (approved):
LGTM! thanks for the fix!

PR #392, opened 2023-03-24
Add Translation Finetuning Example with T5

alexandremuzio:
This PR has 2 changes:

1. Adds `examples/ppo_translation_t5.py`. Which shows an example of finetuning a T5 model for sequence, optimizing for COMET which is a neural metric trained to correlate with Human judgements. Should work for any HF translation model as well.

2. Small change to the `trlx/trainer/accelerate_base_trainer.py` to allow for agreggated metrics to be logged during training. This is needed when reporting corpus level metrics for evaluation.

### DIFF
--- a/examples/ppo_translation_t5.py
+++ b/examples/ppo_translation_t5.py
@@ -21,8 +21,8 @@ from trlx.data.configs import (
 from trlx.models.modeling_ppo import PPOConfig
 
 try:
-    import evaluate
     import comet
+    import evaluate
 
     if comet.__version__ != "1.1.3":
         raise ImportError
--- a/examples/ppo_translation_t5.py
+++ b/examples/ppo_translation_t5.py
@@ -22,9 +22,13 @@ from trlx.models.modeling_ppo import PPOConfig
 
 try:
     import evaluate
+    import comet
+
+    if comet.__version__ != "1.1.3":
+        raise ImportError
 except ImportError:
     raise ImportError(
-        "To run this example, please install the `evaluate` and `nltk` packages" "by running `pip install evaluate`"
+        "To run this example, please install `evaluate`, `nltk` and `comet==1.1.3` packages by running `pip install evaluate unbabel-comet==1.1.3`"
     )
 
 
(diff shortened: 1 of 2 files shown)


examples/ppo_translation_t5.py
```diff
@@ -0,0 +1,197 @@
...
+            "max_new_tokens": 100,
+            "do_sample": False,
+            "num_beams": 4,
+            "temperature": 1.0,
+        },
+    ),
+)
+
+# COMET is the metric we are optimizng for
+comet_metric = evaluate.load("comet", "wmt20-comet-da", progress_bar=False)
```
jon-tow, line 98: I'm having trouble loading comet and not sure if this is a general issue. Is `wmt20-comet-da` stored in a private repo? See repro here: https://colab.research.google.com/drive/1Du3-sIff2mz-34I62NP2pjmRRezTPGtH?usp=sharing

alexandremuzio: Huum, I was able to reproduce your issue as well. It seems the issue with with the newest `comet` package. If you install it with `unbabel-comet==1.1.3`, my code works. With the newest version of `unbabel-comet==2.0.0` which is the one you installed, they removed the `wmt20-comet-da` model and replaced it with a newer performing model called `Unbabel/wmt22-comet-da`.

Since I've only ran with the `wmt20-comet-da`, I think it would be best to stick with the `unbabel-comet==1.1.3` and use `wmt20-comet-da`. Should I add a version check for `unbabel-comet` in the `ppo_translation_t5.py` file?

alexandremuzio: Thanks and done!

jon-tow:
Hi, @alexandremuzio! What sort of eval performance were you getting with this? The default settings implemented in the example lead to divergence in reward/comet/bleu etc. See here: https://api.wandb.ai/links/carperai/5wakyx9f

alexandremuzio:
Interesting, I'll take a look at what might be going wrong

LouisCastricato:
Lets move this to draft for the time being.

alexandremuzio:
Hi @LouisCastricato  and @jon-tow. I had forgotten to update the hparams and here is my latest run: https://api.wandb.ai/links/alexmuzio/84c95q2h

jon-tow (approved):
Looks much better with the improved hparams 👍 Thank you!

LouisCastricato:
If Jon is approving I'm ok to merge.

PR #394, opened 2023-03-26
Add support for batch gradient accumulation

eluzhnica:
This adds support for accumulating gradients across many batches for PPO.

If you're constrained on the number of GPUs available, the effective batch size is going to be likely a lot smaller than the default configurations that have been tested here. Lowering the learning rate, in that case, should solve most of those. However, I did run into more instabilities where the reward was collapsing for small effective batch sizes (i.e 2) when training gpt-j.

This PR allows us to set the gradient accumulation step to accumulate over that many batches, this way we can get comparable behavior to training across many gpus.

I've tested across many runs on GPT-j (although only on 1 gpu, +1 gpu for reward model) and it does seem to have solved those instabilities.

### Implementation notes:

The most straightforward implementation would have been to simply flip the looping order - iterate n_updates_per_batch over all minibatches in a rollout, but it might not be the right PPO implementation. Ran some tests on GPT-j and that works just fine. Also looking at the implementation of ppo in OpenAI Baselines suggests that might actually be right (https://github.com/openai/baselines/blob/master/baselines/ppo2/ppo2.py#L158). However, as this would constitute a behavior change, I chose to implement the smallest change possible, without altering past behavior. Plus, I'm sure this was probably considered.

This change does modify the behavior if gradient accumulation steps >1 was used for multiple n_updates_per_batch. However, accumulating gradients over the same batch isn't very sound.

If there is consensus on implementing the more straightforward method mentioned above, I can update the PR accordingly and run a few more tests.

### DIFF
--- a/trlx/trainer/accelerate_base_trainer.py
+++ b/trlx/trainer/accelerate_base_trainer.py
@@ -1,7 +1,9 @@
+import contextlib
 import json
 import os
 import sys
 from abc import abstractmethod
+from contextlib import contextmanager
 from time import time
 from typing import Dict, List, Optional, Tuple
 
@@ -422,6 +424,18 @@ class AccelerateRLTrainer(BaseRLTrainer):
         self.nth_evaluation += 1
         return stats
 
+    @contextmanager
+    def _accumulate(self):
+        if (
+            (self.iter_count + 1) % self.accelerator.gradient_accumulation_steps == 0
+            or self.iter_count + 1 == self.config.train.total_steps
+        ):
+            context = contextlib.nullcontext
+        else:
+            context = self.accelerator.no_sync
+        with context(self.model):
+            yield
+
     def learn(self):  # noqa: C901
         """
         Samples batches from `self.store`, updates model and periodically evaluates it on `self.eval_dataloader`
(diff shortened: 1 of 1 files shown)


trlx/trainer/accelerate_base_trainer.py
```diff
@@ -467,78 +481,89 @@ def learn(self):  # noqa: C901
         # For each epoch
         for _ in range(self.config.train.epochs):
             # For each batch
-            for batch in self.train_dataloader:
-                # For each update per batch
+            dataloader_iterator = iter(self.train_dataloader)
+            for initial_batch in dataloader_iterator:
+                batches = [initial_batch]
+                if self.accelerator.gradient_accumulation_steps > 1 and self.n_updates_per_batch > 1:
```
eluzhnica, line 487: this check could be skipped but if someone was using a custom dynamic dataloader for rollouts in ppo, it would change the behavior to load up more batches in memory. I don't know how far we want to go to support those cases though. 

and ofc it still changes the behavior if they were using rgadient_accumulation_steps > 1, but I assume nobody was doing that for n_updates_per_batch > 1 since I don't think there's a point in doing that (accumulating gradients over the same batch multiple times)

eluzhnica:
@Dahoas I just noticed you do have this: https://github.com/CarperAI/trlx/pull/364/files which achieves the same thing and is probably more principled anyways. I can close this MR and just add a commit there to avoid the gradient synchronization

eluzhnica:
closing to do this instead https://github.com/CarperAI/trlx/pull/364/files

PR #395, opened 2023-03-27
[fix] update pairwise dataloader.

Chen9154:
In forward() of reward_model.py (Line 62), if "chosen" and "rejected" are exactly the same, "inference" would turn to True, which shouldn't happen during the training procedure. However in class PairwiseDataset, "chosen" and "rejected" could be the same after truncation (this would be easily happen when prompts/posts are longer than max_length and we set padding_side = 'right'). So we filter out those cases from training data.


jon-tow (changes requested):
Hi, @Chen9154! Thanks for catching this edge case; looks good to me 👍 

Would you be able to also make this same change to the test file for sake of completeness, please?
https://github.com/CarperAI/trlx/blob/9fd1f0a7e81d13e8bb04bad96790569b2eb0cfb1/examples/summarize_rlhf/reward_model/gptj_reward_test.py#L40-L65

Chen9154:
@jon-tow Thanks for the review! I have also made the same change to the test file.

Issue #373, opened 2023-03-16
PPO training fails with NCCL timeout when running on larger models 


agave233:
Hello,

I have successfully run the code summarize_rlhf with small SFT and RM models (bloom1b). However, when I try to run the larger model (7B), **the timeout error is raised,** which is a similar problem as stated in this issue #319. But I can not find a solution.

My environment:
trlx version: 0.5.0
accelerate: 0.17.1
torch version: 1.13.1

The error is as follows:
 ```
rollout 15 / 16]:  94%|█████████▍| 15/16 [00:25<00:01,  1.81s/it]
[rollout 15 / 16]:  94%|█████████▍| 15/16 [00:27<00:01,  1.81s/it][E ProcessGroupNCCL.cpp:737] [Rank 2] Watchdog caught collective operation timeout: WorkNCCL(SeqNum=2059, OpType=ALLREDUCE, Timeout(ms)=1800000) ran for 1800283 milliseconds before timing out.
╭───────────────────── Traceback (most recent call last) ──────────────────────╮
│ /root/paddlejob/workspace/env_run/trlx_ppo_bloom_new.py:197 in <module>      │
│                                                                              │
│   194 │   │   post_summary_dict[val_prompts[i]] = val_summaries[i]           │
│   195 │                                                                      │
│   196 │   reward_fn = create_reward_fn()                                     │
│ ❱ 197 │   trainer = trlx.train(                                              │
│   198 │   │   reward_fn=reward_fn,                                           │
│   199 │   │   prompts=train_prompts,                                         │
│   200 │   │   eval_prompts=val_prompts[0:1000],  # sampling 1000 validation  │
│                                                                              │
│ /root/paddlejob/workspace/env_run/tools/trlx/trlx/trlx.py:123 in train       │
│                                                                              │
│   120 │   eval_pipeline = get_pipeline(config.train.pipeline)(eval_prompts,  │
│   121 │   trainer.add_eval_pipeline(eval_pipeline)                           │
│   122 │                                                                      │
│ ❱ 123 │   trainer.learn()                                                    │
│   124 │   return trainer                                                     │
│   125                                                                        │
│                                                                              │
│ /root/paddlejob/workspace/env_run/tools/trlx/trlx/trainer/accelerate_base_tr │
│ ainer.py:478 in learn                                                        │
│                                                                              │
│   475 │   │   │   │   │   # multiple gradient updates on the same batch of d │
[...]

PR #401, opened 2023-03-28
[fix] Preserve `<eos>` token and in-place it after trimming

### LINKED ISSUE #399: PPO training stuck on `make_experience`
### 🐛 Describe the bug

The following script would get stuck indefinitely on the second rollout collection:
```python
import trlx
from trlx.data.default_configs import default_ppo_config

def main():
    config = default_ppo_config()
    config.train.seq_length = 1024
    config.model.model_path = "gpt2"
    config.method.gen_kwargs["top_k"] = 1

    prompts = ["*" * 256] * 1024

    trlx.train(
        reward_fn=lambda samples, **kwargs: [0] * len(samples),
        prompts=prompts,
        eval_prompts=prompts,
        config=config,
        stop_sequences=["*"],
    )

if __name__ == "__main__":
    main()
```
```bash
accelerate launch --num_processes 1 --config_file configs/accelerate/zero2-bf16.yaml examples/ppo_stuck.py
```

latest @ https://github.com/CarperAI/trlx/tree/114998b8171e7ce8d543b2102efe8a612a01ff9f

from https://github.com/CarperAI/trlx/blob/114998b8171e7ce8d543b2102efe8a612a01ff9f/requirements.txt

reciprocated:
This PR fixes #399 and #373 by keeping `<|endoftext|>` token if it was outputted by the model (previously it was treated as if it was one of `stop_sequences`) and replacing trimming after-and-including `stop_sequences` with `<|endoftext|>`, instead of simply discarding them. This lets PPO's `make_experience` never filter any empty outputs, or in the worst case, stuck and timeout.

https://wandb.ai/sorry/trlx-references/reports/fix-preserve-eos-token-v-main--VmlldzozOTE1MDU2

### DIFF
--- a/trlx/trainer/accelerate_ppo_trainer.py
+++ b/trlx/trainer/accelerate_ppo_trainer.py
@@ -302,7 +302,7 @@ class AcceleratePPOTrainer(AccelerateRLTrainer):
 
             if self.accelerator.is_main_process:
                 all_str_samples, all_str_prompts, all_str_outputs = self.decode(
-                    gathered_prompts, gathered_samples, gathered_prompt_sizes
+                    gathered_prompts, gathered_samples, gathered_prompt_sizes, append_eos_token=True
                 )
 
                 exp_score_time = time()
@@ -327,7 +327,7 @@ class AcceleratePPOTrainer(AccelerateRLTrainer):
             else:
                 scores = all_scores[0].clone().detach()
 
-            str_samples, str_prompts, str_outputs = self.decode(prompt_tensors, samples)
+            str_samples, str_prompts, str_outputs = self.decode(prompt_tensors, samples, append_eos_token=True)
 
             # Pad the sample outputs
             outputs = self.tokenizer(str_outputs).input_ids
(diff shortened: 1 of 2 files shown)


trlx/trainer/accelerate_ppo_trainer.py
```diff
@@ -445,7 +445,7 @@ def make_experience(self, num_rollouts: int = 1024, iter_count: int = 0):  # noq
             sample_outputs = sample_outputs.cpu()
             values = values.cpu()[:, :-1]
 
-            ends = start + attention_mask[:, start:].sum(1)
+            ends = start + attention_mask[:, start:].sum(1) + 1
```
jon-tow, line 448: Nit: Maybe add a comment saying something like "add 1 to account for appended eos  token" since the decoding is a few blocks up

jon-tow (approved):
The dreaded filter is gone hehe Great work! I left a small nit but feel free to ignore and merge whenever you'd like.

(Also, first time seeing your benchmark util in practice. Love it 👍)

PR #402, opened 2023-03-29
[fix] add `device` argument to the call of `modeling_opt._make_causal_mask`

shionhonda:
This PR fixes a bug in the call of `modeling_opt._make_causal_mask()` in `OPTModelBranch`. This private method requires `device` argument, as found in:
https://github.com/huggingface/transformers/blob/b29fd6971d9cd6ba2a824628effe243f543b8f61/src/transformers/models/opt/modeling_opt.py#L68-L70

### DIFF
--- a/trlx/models/modeling_ppo.py
+++ b/trlx/models/modeling_ppo.py
@@ -597,8 +597,9 @@ class OPTModelBranch(ModelBranch):
             combined_attention_mask = modeling_opt._make_causal_mask(
                 input_shape,
                 hidden_states.dtype,
+                device=hidden_states.device,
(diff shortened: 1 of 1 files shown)


jon-tow (approved):
Thanks for the quick fix, @shionhonda!

jon-tow:
This update is only in the latest `main` branch, so let's wait until a patch release is available to merge this. For now, try to use an earlier version of `transformers`, if possible 🙏

shionhonda:
Ah, I see! I was looking at the latest main branch. This change in the API happened very recently.

PR #404, opened 2023-03-30
Nemo ILQL fixes

cat-state:
Update the nemo ilql make_experience function to reuse the accelerate one. I forgot to update it with the hypothesis test additions. Also the yaml -> python config change accidently set the default beta for generations for nemo_ilql_sentiments to 4, when previously it was 2, as well as changed the temperature. This restores the previous results (although it is task specific tuning for ILQL)

https://wandb.ai/uwu1/trlxnemo/reports/make-experience-change--VmlldzozOTI3MDU5


jon-tow (approved):
Nice catch and cleanup 👍  LGTM!

PR #406, opened 2023-03-31
Add some synchronization to the db download in the simulacra example

dakinggg:
Hi, I'm running the simulacra example just to make sure that our distributed setup works with the TRLX examples, and it seems there is a bit of a race condition with respect to downloading the db, so I added some logic for only local rank 0 to download the db, and a barrier after the download. Let me know if there is another way you prefer to handle this (e.g. we could use sleep for a less reliable solution if you don't want to introduce an accelerate dependency in this example). Thanks!

Screenshot of single and multi node runs
Screen Shot 2023-03-31 at 11 51 57 AM

### DIFF
--- a/examples/simulacra.py
+++ b/examples/simulacra.py
@@ -1,6 +1,7 @@
 # Optimize prompts by training on prompts-ratings pairings dataset
 # taken from https://github.com/JD-P/simulacra-aesthetic-captions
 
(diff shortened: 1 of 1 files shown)


jon-tow (approved):
`accelerate` is pretty much a hard dependency with trlx so no issue using it here. Thanks for the fix!

PR #409, opened 2023-04-01
set torch dependency to version 2.0.0 for CUDA in installation instru…

cauyxy:
If there is already a version of Torch installed in your environment, using the same commands won't overwrite it. This can cause some code that uses Torch 2.0 to throw errors. Therefore, it is necessary to explicitly specify the Torch version and also consider upgrading the version of NCCL.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ The following RL algorithms are currently implemented:
 ```bash
 git clone https://github.com/CarperAI/trlx.git
 cd trlx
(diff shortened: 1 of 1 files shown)


cauyxy:
such as torch.distributed.scatter、torch.distributed.ReduceOp.AVG

reciprocated (approved):
Right, I think that it's good to upfront these requirements you've mentioned, thanks!

PR #414, opened 2023-04-03
Set pad_token for all tokenizers in tests

cat-state:
Tests were breaking due to one of the `self._create_inputs` in the model tests not doing this


jon-tow (approved):
Strange that `tokenizer.pad_token = tokenizer.eos_token`  didn't silence the errors.
Looks good to me though 😁 Thanks, Aman!

PR #417, opened 2023-04-04
Convert tensors in the stats dict into scalars

ZHAOTING:
The values of some items in the stats dictionary are tensors in the current code, and it prevents a tensorboard accelerate tracker from logging them.
This PR converts those tensors into scalars via `.item()` and makes the tensorboard tracker behave as expected.


jon-tow (approved):
Thanks, @ZHAOTING!

PR #419, opened 2023-04-04
Create Example training scripts to run in Stability cluster

alexandremuzio:
Example scripts that run `examples/ilql_sentiments.py` on stability AWS cluster. Should work for other examples
Supports multi-node training.

Based on script from @reciprocated


alexandremuzio:
@PhungVanDuy , when you have some time, could you give this script a try?

reciprocated (approved):
Thanks Alexandre! LGTM

PR #420, opened 2023-04-05
Ensure trailing EOS token is added correctly for shorter generated outputs

mikljohansson:
Ensure trailing EOS token is added also when the generated output contains trailing padding as is often the case when doing batch generation. When doing batched generations, some generated samples will end earlier than other samples. The samples that ended earlier will (in the case of T5/MT5) contain trailing `<pad>` tokens. The `sample[-1]` test to check if the last token of the sample is `<eos>` only works when the `<eos>` token comes last, which is only true for the longest of the generated samples.

### DIFF
--- a/trlx/trainer/accelerate_base_trainer.py
+++ b/trlx/trainer/accelerate_base_trainer.py
@@ -219,9 +219,11 @@ class AccelerateRLTrainer(BaseRLTrainer):
 
             # Recover the last <eos> if it was present in the original sample
             # or add one if it was trimmed with `self.stop_sequences`.
-            # Only in cases when a generation ended due to `max_new_tokens` exhaustion,
-            # <eos> token would not be present in the original sample
-            if append_eos_token and (trimmed or self.tokenizer.eos_token_id in sample):
+            # When a generation ended due to `max_new_tokens` exhaustion,
+            # only then <pad> or <eos> token would not be present in the original sample at the end
+            if append_eos_token and (
+                trimmed or sample[-1] == self.tokenizer.eos_token_id or sample[-1] == self.tokenizer.pad_token_id
+            ):
                 str_output += self.tokenizer.eos_token
 
             str_prompts.append(str_prompt)
--- a/trlx/trainer/accelerate_base_trainer.py
+++ b/trlx/trainer/accelerate_base_trainer.py
@@ -221,7 +221,7 @@ class AccelerateRLTrainer(BaseRLTrainer):
             # or add one if it was trimmed with `self.stop_sequences`.
             # Only in cases when a generation ended due to `max_new_tokens` exhaustion,
             # <eos> token would not be present in the original sample
-            if append_eos_token and (trimmed or sample[-1] == self.tokenizer.eos_token_id):
+            if append_eos_token and (trimmed or self.tokenizer.eos_token_id in sample):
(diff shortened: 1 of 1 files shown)


trlx/trainer/accelerate_base_trainer.py
```diff
@@ -210,7 +210,7 @@ def decode(
             # or add one if it was trimmed with `self.stop_sequences`.
             # Only in cases when a generation ended due to `max_new_tokens` exhaustion,
             # <eos> token would not be present in the original sample
-            if append_eos_token and (trimmed or sample[-1] == self.tokenizer.eos_token_id):
```
reciprocated, line 213: Do you think maybe it's better to check for `pad_token` as the last token instead, to account for the case when `eos_token` might be present in the prompt?

reciprocated (commented):
Thanks for spotting this! It weren't working as expected in the case when `pad_token != eos_token`.

trlx/trainer/accelerate_base_trainer.py
mikljohansson: Hmm, I'm not sure honestly. What I see is that I get batches like this (showing it with text instead of int tokens)

```
prompts = ["<pad>Some prompt 1", "Another prompt 2"]
samples = ["<s>Completion from T5 with prompt 1</s><pad><pad><pad>", "<s>A longer completion from T5 with prompt 2</s>"]
```

The original code created outputs like below (notice that only one of the generations have the </s>). The <pad> and <s> tokens are stripped by `tokenizer.decode(.., skip_special_tokens=True)`

```
samples = ["Completion from T5 and prompt 1", "A longer completion from T5 and prompt 2</s>"]
```

The actual tensors with tokens look like for example below (0 is used for both <s> (BOS) and <pad>, 1 is used for </s> (EOS)). Notice also that one of the prompt tensors is left-padded, which I believe might also be incorrect for an encoder-decoder model (I'm trying to debug this next)

```
prompts = [[0, 123, 456, ...], [312, 423, 675, ...]]
samples = [[0, 523, 341, ..., 1, 0, 0, 0], [0, 623, 523, ..., 1]]
```

Another possible issue I'm trying to investigate is that at a later stage the output/reward tensors have two </s> tokens, and the reward seems to be added to the last </s> token in `AcceleratePPOTrainer.make_experience`. I'm assuming the reward should be added to the first </s> (eos token) right?

For example like this

```
outputs = [[123,456, 1, 1], ...]
rewards = [[0, 0, 0, 5.25], ...]
```

reciprocated: Regarding the trailing EOS token issue, it seems to me that a proper check, which would respect both decoder-only and encoder-decoder models, here would be `sample[-1] == eos_token_id or sample[-1] == pad_token_id`, because `sample` in decoder-only case also contains the prompt, which may be just an `eos_token`. I am also glad you've opened a PR, which addresses the second issue you've mentioned here, and I will try to debug the third one as well. This is very valuable!

mikljohansson: Ok, that makes sense. Thanks for looking into this!

reciprocated (approved):
Okay, I'm going to merge the version I suggested here, you seem to have force pushed the the same commit instead

Issue #340, opened 2023-02-28
train trlx_gptj_text_summarization.py  post_summary_dict key error


jack-Dong:
### 🐛 Describe the bug

```
[losses/total_loss: 0.34 | losses/policy_loss: 0.31 | losses/value_loss: 0.14]:   4%|████▍                                                                                                         | 256/6400 [14:08<1:23:11,  1.23it/s][RANK 0] Collecting rollouts
[rollout 16 / 128]:  12%|█████████████████████▊                                                                                                                                                        | 16/128 [00:10<01:12,  1.54it/s]Traceback (most recent call last):
  File "trlx_gptj_text_summarization.py", line 123, in <module>
    trainer = trlx.train(
  File "/cloudgpfs/workspace/dongshizhou/py_porjetcs/trlx/trlx/trlx.py", line 117, in train
    trainer.learn()
  File "/cloudgpfs/workspace/dongshizhou/py_porjetcs/trlx/trlx/trainer/accelerate_base_trainer.py", line 541, in learn
    self.post_epoch_callback()
  File "/cloudgpfs/workspace/dongshizhou/py_porjetcs/trlx/trlx/trainer/accelerate_ppo_trainer.py", line 227, in post_epoch_callback
    self.make_experience(self.config.method.num_rollouts, self.iter_count)
  File "/cloudgpfs/workspace/dongshizhou/py_porjetcs/trlx/trlx/trainer/accelerate_ppo_trainer.py", line 311, in make_experience
    self.reward_fn(
  File "trlx_gptj_text_summarization.py", line 87, in reward_fn
    original_samples = [text + post_summary_dict[text.strip()] for text in original_samples]
  File "trlx_gptj_text_summarization.py", line 87, in <listcomp>
    original_samples = [text + post_summary_dict[text.strip()] for text in original_samples]
[...]

Issue #301, opened 2023-02-09
Pass extra information for the reward function with every sample.


JulesGM:
### 🚀 The feature, motivation, and pitch

There are multiple situations where having more information about the sample than just the string would be good with the reward function. A very dumb example is if you want spanish sentences to be shorter and english sentences to be longer. You would have to pass the language of the sample. 

The only option right now is to use a hashmap, however, this has disadvantages. 
 - It requires keeping the whole dataset in memory
 - Because of tokenization artifacts, the access may sometimes break
 - It breaks expectations of how this should work. People likely expect just to be able to pass information with the dataset object.

As library a builder it's hard to predict all the uses that the library will have, so I suggest just being able to pass a dict per sample with the dataloader, and to pass a list of dicts to the reward function.

maxreciprocate:
Resolved with https://github.com/CarperAI/trlx/pull/421

Issue #297, opened 2023-02-09
(perhaps) data preparation bug in RLHF


xwwwwww:
### 🐛 Describe the bug

I have several problems about data preparations while running `summarize_rlhf`.

1. In `get_prompt_dataset` https://github.com/CarperAI/trlx/blob/f115eeaa3cfd2c997a345b8891b5f9427f1a08ee/examples/summarize_rlhf/trlx_gptj_text_summarization.py#L58
Current implementations will first truncate the original prompts with `max_length-5`. However, I notice that the tokenized results of the prompts will change after appending `\nTL;DR:` in some cases, which then lead to the truncation of the suffix of `\nTL;DR:` and finally lead to keyerror in https://github.com/CarperAI/trlx/blob/f115eeaa3cfd2c997a345b8891b5f9427f1a08ee/examples/summarize_rlhf/trlx_gptj_text_summarization.py#L83

2. Even after I fix this truncation bug in `get_prompt_dataset`, current implementations will still raise keyerror in https://github.com/CarperAI/trlx/blob/f115eeaa3cfd2c997a345b8891b5f9427f1a08ee/examples/summarize_rlhf/trlx_gptj_text_summarization.py#L83 during PPO training.

### Which trlX version are you using?

main

### Additional system and package information

transformers==4.26.0

LouisCastricato:
@PhungVanDuy

LouisCastricato:
Thanks, we will investigate

LouisCastricato:
We should freeze our transformers version maybe... just update it every major TRLX release.

PhungVanDuy:
> 

Yes, even with the torch version, accelerate,... some people raised errors when they ran on different versions with us.

xwwwwww:
>  I suggested you change the function with this style: https://github.com/CarperAI/trlx/blob/main/examples/summarize_daily_cnn/t5_summarize_daily_cnn.py#L25, to make sure the KeyError not happened again. I will make a new version of this examples soon to prevent this error happened again.

Could you please provide more detailed instructions on how to fix the above issue? I didn't see clear difference between https://github.com/CarperAI/trlx/blob/main/examples/summarize_daily_cnn/t5_summarize_daily_cnn.py#L25 and current implementations of `summarize_rlhf` :)

xwwwwww:
by the way, I have tried `transformers=4.25.1` and it still raises KeyError

xwwwwww:
[...]

PR #421, opened 2023-04-05
[feat] Support tying metadata to each prompt

reciprocated:
Fix of #297, #301, #340

### DIFF
--- a/trlx/pipeline/offline_pipeline.py
+++ b/trlx/pipeline/offline_pipeline.py
@@ -1,5 +1,5 @@
 from dataclasses import dataclass
-from typing import Iterable, List, Tuple, Union
+from typing import Any, Dict, Iterable, List, Tuple, Union
 
 import torch
 from torch.nn.utils.rnn import pad_sequence
@@ -103,8 +103,16 @@ class PromptPipeline(BasePipeline):
     Tokenizes prompts, unless they are already tokenized, and truncates them to `max_prompt_length` from the right
     """
 
-    def __init__(self, prompts: List[str], max_prompt_length: int, tokenizer: PreTrainedTokenizer):
+    def __init__(
+        self, prompts: Union[Dict[str, Any], List[str]], max_prompt_length: int, tokenizer: PreTrainedTokenizer
+    ):
         super().__init__()
+
+        if isinstance(prompts[0], dict):
+            metadata = prompts
+            prompts = [x.pop("prompt") for x in metadata]
+        else:
+            metadata = [{}] * len(prompts)
 
         model_inputs = tokenizer(
             prompts, truncation=True, padding=False, max_length=max_prompt_length, add_special_tokens=False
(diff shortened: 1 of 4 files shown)


trlx/pipeline/offline_pipeline.py
```diff
@@ -125,7 +134,15 @@ def __len__(self) -> int:
         return len(self.prompts)
 
     def create_loader(self, batch_size: int, shuffle=False) -> DataLoader:
-        collate_fn = DataCollatorWithPadding(self.tokenizer) if self.tokenizer else torch.vstack
+        def collate_fn(xs):
+            out = self.tokenizer.pad([{"input_ids": x["input_ids"]} for x in xs], return_tensors="pt")
```
jon-tow, line 138: Is it safe to assume `self.tokenizer` will always be available? Not sure if the previous check `if self.tokenizer else torch.vstack` is just stale code?

trlx/pipeline/offline_pipeline.py
```diff
@@ -103,9 +103,17 @@ class PromptPipeline(BasePipeline):
...
 
-    def __init__(self, prompts: List[str], max_prompt_length: int, tokenizer: PreTrainedTokenizer):
+    def __init__(
+        self, prompts: Union[Dict[str, Any], List[str]], max_prompt_length: int, tokenizer: PreTrainedTokenizer
+    ):
         super().__init__()
 
+        if isinstance(prompts[0], dict):
+            metadata = prompts
+            prompts = [x.pop("prompt") for x in metadata]
```
jon-tow, line 113: In `PromptPipeline`'s constructor docstring, we should document that `"prompt"` is an expected field of `metadata` when `prompts` consists of dictionaries.

jon-tow (changes requested):
Looks great, Max! I'm definitely going to steal your `gather_dict` util for personal code 😄
I've left two comments for feedback whenever you get the chance 🙏

trlx/trainer/accelerate_ppo_trainer.py
```diff
@@ -282,7 +282,7 @@ def make_experience(self, num_rollouts: int = 1024, iter_count: int = 0):  # noq
             exp_generate_time = time()
 
             # Generate samples from the language model (similar to using HuggingFace `generate` method)
-            samples = self.generate(**batch)
+            samples = self.generate(batch["input_ids"], batch["attention_mask"])
```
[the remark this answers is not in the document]
LouisCastricato: Why don't we depreciate prompt batch then.

trlx/pipeline/offline_pipeline.py
reciprocated: It was a stale code and the single example which didn't use tokenizer at the time (ilql_randomwalks) was also converted to use tokenizer to drop these checks, out of which it is the last, for consistency; right now tokenizer is required in almost every other piece of code, so I'm not sure if that feature is missed.

PR #423, opened 2023-04-05
fix(CI): use pinned deps for CI testing

jon-tow:
Use the pinned dependencies in `requirements.txt` for unit testing in CI re discussion in the recent trlx meeting.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,7 +23,13 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install -e ".[dev,bnb]"
+        pip install torch
+        pip install -r requirements.txt
+        # Install extras
+        # [bnb]
+        pip install bitsandbytes 
+        # [dev]
+        pip install black hypothesis isort flake8 pre-commit pytest pytest-cov
 
     - name: Lint with flake8
       run: |
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -23,7 +23,6 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -23,7 +23,13 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install -e ".[dev,bnb]"
+        pip install torch
```
reciprocated, line 26: This would get later overwritten by
```
torch==2.0.0+cu117
```
from requirements.txt

Build log:
```
Collecting torch==2.0.0+cu117
  Downloading https://download.pytorch.org/whl/cu117/torch-2.0.0%2Bcu117-cp38-cp38-linux_x86_64.whl (1843.9 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 1.8/1.8 GB 699.2 kB/s eta 0:00:00
```

reciprocated (approved):
Thanks Jon! I've checked dependencies used on CI to be the same as in requirements, and since the build passes that means it's all good!

.github/workflows/build.yml
jon-tow: Oh huh. Running `pip install -r requirements.txt` locally I get:
```
ERROR: Could not find a version that satisfies the requirement torch==2.0.0+cu117 (from versions: 1.11.0, 1.12.0, 1.12.1, 1.13.0, 1.13.1, 2.0.0)
ERROR: No matching distribution found for torch==2.0.0+cu117
```
I guess the CI hosts have CUDA/GPU support? I'll remove the lone `pip install torch`!

jon-tow:
Thanks for the review Max!

PR #429, opened 2023-04-10
fix(base_trainer): gather weights in `save_pretrained` under zero3

reciprocated:
This PR lets users optionally choose whether to save optimizer state on each and a final checkpoint or only the model weights via `trainer.save_pretrained` (new default). Also this PR removes two back to back calls to `trainer.evaluate`, which would happen on the same `iter_count` in the case `total_steps` is divisible by `config.train.eval_interval`. And it fixes `trainer.save_pretrained`, in which under zero3, weights weren't gathered previously into the `state_dict`.

https://wandb.ai/sorry/trlx-references/reports/fix-save-pretrained-zero3-v-main--Vmlldzo0MDI1ODk1

### DIFF
--- a/trlx/data/configs.py
+++ b/trlx/data/configs.py
@@ -220,7 +220,7 @@ class TrainConfig:
     checkpoint_dir: str = "ckpts"
     rollout_logging_dir: Optional[str] = None
     save_best: bool = True
-    save_optimizer: bool = False
+    save_optimizer: bool = True
 
     tracker: Optional[str] = "wandb"
     logging_dir: Optional[str] = None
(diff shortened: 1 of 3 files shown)


jon-tow (commented):
Thanks for the gather fix and for catching those final step conditions (`self.iter_count >= self.total_steps`)!

One issue with making `save_optimizer = False` the default is that you can't resume training ( even though proper checkpointing from say `deepspeed` ZeRO states is currently WIP). Luckily, we haven't trained on very large datasets (> 10M samples) for this to be much of an issue. Do you foresee this being an issue? I sort of feel like the expected behavior is to save optimizer states.

jon-tow (approved):
Thanks, @reciprocated!

PR #433, opened 2023-04-13
Avoid a few off-by-one issues which PAD and EOS tokens in the generated sequences

mikljohansson:
* Calculate KL divergence also for the logprobs from the BOS token for seq2seq models by setting `attention_mask[:, 0] = 1` similar to how is done a bit further up in the code
* Avoid including the first PAD token in all_values/all_logprobs. The EOS seems to already be included in the `attention_mask` so there should no need to add + 1 to `ends` I believe, as that also include a PAD token into the logprobs when I step through the program in a debugger. 
* Use the generated tensors from seq2seq model directly, instead of decoding and encoding them which ended adding duplicate EOS tokens to the end

### DIFF
--- a/trlx/trainer/accelerate_ppo_trainer.py
+++ b/trlx/trainer/accelerate_ppo_trainer.py
@@ -325,14 +325,12 @@ class AcceleratePPOTrainer(AccelerateRLTrainer):
             else:
                 scores = all_scores[0].clone().detach()
 
-            str_samples, str_prompts, str_outputs = self.decode(prompt_tensors, samples, append_eos_token=True)
-
             # Pad the sample outputs
-            outputs = self.tokenizer(str_outputs).input_ids
             if self.config.model.model_arch_type == "seq2seq":
-                # add <pad> to the start of the output
-                for i in range(len(outputs)):
-                    outputs[i] = [self.tokenizer.pad_token_id] + outputs[i]
+                outputs = samples.detach().cpu()
+            else:
+                _, _, str_outputs = self.decode(prompt_tensors, samples, append_eos_token=True)
+                outputs = self.tokenizer(str_outputs).input_ids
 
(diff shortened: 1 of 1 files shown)


trlx/trainer/accelerate_ppo_trainer.py
```diff
@@ -446,7 +445,7 @@ def make_experience(self, num_rollouts: int = 1024, iter_count: int = 0):  # noq
             # Get the logprobs and values, for tokens that are not padding,
             # from the start of the prompt up to the <eos> token, while also including the latter
             # (these are taken from the student model and not the reference model)
-            ends = start + attention_mask[:, start:].sum(1) + 1
+            ends = start + attention_mask[:, start:].sum(1)
```
reciprocated, line 448: This will be correct (with the correction to `attention_mask` above), however not for tokenizers which have `<eos> == <pad>`. By construction `attention_mask != <pad>`, so this additional +1 index was getting the last `<eos>` token, which would be treated as padding otherwise.

trlx/trainer/accelerate_ppo_trainer.py
```diff
@@ -325,14 +325,12 @@ def make_experience(self, num_rollouts: int = 1024, iter_count: int = 0):  # noq
             else:
                 scores = all_scores[0].clone().detach()
 
-            str_samples, str_prompts, str_outputs = self.decode(prompt_tensors, samples, append_eos_token=True)
```
reciprocated, line 328: I don't find (from the current main, and from previously merged PRs) any superflous `<eos>` tokens added through the `decode` call here. But probably skipping it is excessive, since it still cuts `stop_sequences`

maxreciprocate (changes requested):
Off-by-one issues are very annoying, thank you for going through them! However there is one place here where such issue might be fixed but also reintroduced for a different tokenizer

PR #435, opened 2023-04-14
fix(offline_pipeline): ILQL negative indexing under truncation

reciprocated:
This PR fixes a bug under which it was possible to run into negative indices with tokenizer's left truncation

https://wandb.ai/sorry/trlx/runs/yn59xu9i

### DIFF
--- a/tests/test_pipelines.py
+++ b/tests/test_pipelines.py
@@ -28,7 +28,6 @@ class TestTokenizeDialog(TestCase):
         assert user_dm == DialogMessage(is_output=False, tokens=(self.tokenizer.bos_token_id,))
         assert bot_dm == DialogMessage(is_output=True, tokens=tokenized_response)
 
-
     @given(st.lists(st.text(), max_size=32))
     def test_tokenize_dialogue_single_turn(self, response_words):
         response = " ".join(response_words)  # space seperate to make it multiple tokens
--- a/tests/test_pipelines.py
+++ b/tests/test_pipelines.py
@@ -11,22 +11,18 @@ class TestTokenizeDialog(TestCase):
     def setUp(self):
         self.tokenizer = AutoTokenizer.from_pretrained("gpt2")
 
-    def test_tokenize_dialogue_truncation(self):
-        dialogue = ["will be truncated", "ø" * 1024]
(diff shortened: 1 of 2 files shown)


tests/test_pipelines.py
```diff
@@ -11,6 +11,23 @@ class TestTokenizeDialog(TestCase):
     def setUp(self):
         self.tokenizer = AutoTokenizer.from_pretrained("gpt2")
 
+    def test_tokenize_dialogue_truncation(self):
+        dialogue = ["will be truncated", "ø" * 1024]
```
LouisCastricato, line 15: Should probably use something that is ACII rather than unicode.

trlx/pipeline/offline_pipeline.py
```diff
@@ -34,7 +34,7 @@ def tokenize_dialogue(  # noqa: C901
     if isinstance(dialogue, str):
         bos_token = tokenizer.bos_token or tokenizer.eos_token
         dialogue = [bos_token, dialogue]
-    elif isinstance(dialogue, tuple):
+    elif isinstance(dialogue, Iterable):
```
cat-state, line 37: You could update the type in the signature too

cat-state (approved):
I saw one potential edge case but LGTM to me otherwise

trlx/pipeline/offline_pipeline.py
```diff
@@ -64,9 +64,17 @@ def tokenize_dialogue(  # noqa: C901
         truncated = [DialogMessage(is_output=m.is_output, tokens=m.tokens[::-1]) for m in truncated[::-1]]
 
     # remove empty messages
-    truncated = [t for t in truncated if len(t.tokens) > 0]
+    out = [t for t in truncated if len(t.tokens) > 0]
 
-    return truncated
+    if out[0].is_output:
```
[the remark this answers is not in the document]
reciprocated: it won't, because of enforcing `<eos>` here, output is always non empty
https://github.com/CarperAI/trlx/blob/97f1c4019d67908e55f106796ce8b87df6c75e0d/trlx/pipeline/offline_pipeline.py#L42-L43

PR #441, opened 2023-04-19
fix(ppo_trainer): compute mean KL sequence-wise

### LINKED ISSUE #438: The mean_kl implementation is different from that of openai/lm-human-preference
### 🐛 Describe the bug

Currently, the `mean_kl` used to update `kl_ctl` is calculated from:
https://github.com/CarperAI/trlx/blob/92b68e4d8c5d59e6ba25d12fd9acfe10287be689/trlx/trainer/accelerate_ppo_trainer.py#L437-L438
which is a mean value over each token.
While in [openai/lm-human-preferences](https://github.com/openai/lm-human-preferences), the `mean_kl` is:
```python
            kl = data['logprobs'] - data['ref_logprobs']
            mean_kl = tf.reduce_mean(tf.reduce_sum(kl, axis=1))
```
(https://github.com/openai/lm-human-preferences/blob/bd3775f200676e7c9ed438c50727e7452b1a52c1/lm_human_preferences/train_policy.py#L220-L221)
which is not only the mean value upon each response, but also the same form of the kl used in reward.

Also, in anthropic's paper _Training a Helpful and Harmless Assistant with Reinforcement Learning from Human Feedback_, their kl value could be as large as 25 (as the square root is 5), which is hard to achieve for a token-wise mean kl.
image

I wonder if there is a specific reason why we use the current form of `mean_kl`? Thank you!

Gently ping @Dahoas @reciprocated 

_No response_

_No response_

reciprocated:
This PR fixes #438, specifically:
- Mean KL, which used for statistics and AdaptiveKLController, is now calculated sequence-wise and not token-wise
- Statistics are now averaged across rollouts, instead of just taking them only from the last one
- `exp_*` logging variables are renamed to `rollout_*`, to not confuse them with `sqrt_*`

https://wandb.ai/sorry/trlx/reports/Sequence-wise-v-token-wise-mean-KL--Vmlldzo0MTE0NzUy

https://wandb.ai/sorry/trlx-references/reports/fix-kl-computation-v-main--Vmlldzo0MTE3Nzc4


trlx/trainer/accelerate_ppo_trainer.py
```diff
@@ -435,7 +435,7 @@ def make_experience(self, num_rollouts: int = 1024, iter_count: int = 0):  # noq
                 start = prompt_tensors.shape[1] - 1
 
             log_ratio = (logprobs - ref_logprobs) * attention_mask[:, :-1]
-            self.mean_kl = (log_ratio.exp() - 1 - log_ratio).mean().to(device)
+            mean_kl = (log_ratio.exp() - 1 - log_ratio).sum(1).mean().to(device)
```
Dahoas, line 438: Thanks for fixing this. Do you think it makes sense to add an option to the config allowing us to choose between token wise vs. sequence wise kl? I agree having a kl computation invariant to seq length is good to keep around

trlx/trainer/accelerate_ppo_trainer.py
```diff
@@ -470,18 +470,21 @@ def make_experience(self, num_rollouts: int = 1024, iter_count: int = 0):  # noq
                 )
 
                 rollout_count += 1
-            exp_time = clock.tick()
+
+            if torch.distributed.is_initialized():
```
Dahoas, line 474: Nice

trlx/trainer/accelerate_ppo_trainer.py
```diff
@@ -470,18 +470,21 @@ def make_experience(self, num_rollouts: int = 1024, iter_count: int = 0):  # noq
...
+
             tbar.set_description(f"[rollout {len(ppo_rl_elements)} / {num_rollouts}]")
             tbar.update(min(rollout_count, num_rollouts))
         tbar.close()
 
-        if torch.distributed.is_initialized():
-            torch.distributed.all_reduce(self.mean_kl, torch.distributed.ReduceOp.AVG)
-
-        stats["policy/sqrt_kl"] = torch.sqrt(self.mean_kl).item()
+        stats = {k: sum([xs[k] for xs in accumulated_stats]) / len(accumulated_stats) for k in stats}
```
Dahoas, line 485: Strictly speaking this isn't necessarily correct e.g. if we are recording the max over all local rollouts. However I don't know how to perform the correct reduction without annotating each stat, so this seems fine for now

trlx/trainer/accelerate_ppo_trainer.py
reciprocated: It's easy enough to log both, but for `AdaptiveKLController's purposes I would just stick with one variant which is seq-wise as in lm-human-preferences, since targets are already copied from there and also the respective paper (Ziegler2019 et al.) has target KL value of 8
https://github.com/openai/lm-human-preferences/blob/ec727fde10f1eafb3177e9b0f41a42142e95a2fd/launch.py#L-129-L131

trlx/trainer/accelerate_ppo_trainer.py
reciprocated: There are no maxs/mins in this particular dict, but I see what you are saying. Also I guess you're referring to https://github.com/CarperAI/trlx/blob/9bc08369ca9ec83342c4d7755205dab1a7723006/trlx/trainer/accelerate_base_trainer.py#L544 but it's also pretty easy to fix, perhaps in a separate PR

Dahoas (approved):
Looks good!

lzy37ld:
Thanks for this great work! May you explain a bit why this is sequence-wise KL? I could not connect `'kl = log_ratio.exp() - 1 - log_ratio'` to the real math formulation p(x) logp(x)/q(x).
Gently ping @reciprocated

reciprocated:
@lzy37ld It's sequence-wise KL because of the `.sum(1)` before taking the average, while KL expression itself comes from this blogpost http://joschu.net/blog/kl-approx.html and is commonly used https://github.com/DLR-RM/stable-baselines3/blob/dc09d81f9c07943ddbeac57405d9ae2a31f4d434/stable_baselines3/ppo/ppo.py#L255

lzy37ld:
@reciprocated Thanks! That makes sense! It's really a fantastic implementation!
One more question here: Why [here](https://github.com/CarperAI/trlx/blob/6e655a4a1194eac76d00643c4b5fa56a07384c43/examples/summarize_rlhf/trlx_gptj_text_summarization.py#L150) we would subtract a reward(x,y_original) at the reward_fn? When I looked at the paper, I notice that they only focus on R(x, generated_y)
image

reciprocated:
@lzy37ld It's optional normalization (disabled with setting `delta_reward` to False) https://github.com/CarperAI/trlx/blob/6e655a4a1194eac76d00643c4b5fa56a07384c43/examples/hh/ppo_hh.py#L171
which @PhungVanDuy has found to work better than passing raw reward

PR #455, opened 2023-04-26
Upgrade official released Ray instead of an unstable one.

jovany-wang:
As the Ray-2.4.0 was released with the necessary commits we need, let us upgrade the official released version.


jovany-wang:
duplicate #454

jovany-wang:
@reciprocated  It seems we need to use a stable one.

reciprocated (approved):
Okay nice!
LGTM

A9isha:
Thank you for fixing this!

PR #456, opened 2023-04-26
Correct installation of ray

sanagno:
Ray imports have been changed.


reciprocated:
Hi! Yesterday there was a new Ray release with the feature set we needed, so that we no longer have to use development artifacts. #455 is the PR that corrected the links, could you try if those work for you? It's already merged on main.

PR #460, opened 2023-04-28
fix(trainer): init self.generate_sweep_kwarg at self.__init__

mymusise:
If we do `trainer.evaluate()` before `trainer.learn()`, it will rasie `AttributeError` like:
```
│ ❱ 310 │   │   if self.generate_sweep_kwarg is not None:                                          │
│   311 │   │   │   gen_sweep_arg, gen_sweep_values = self.generate_sweep_kwarg                    │
│   312 │   │   else:                                                                              │
│   313 │   │   │   gen_sweep_values = [None]                                                      │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
AttributeError: 'AccelerateILQLTrainer' object has no attribute 'generate_sweep_kwarg'
```

so, I think it's better to initialize `generate_sweep_kwarg` with the initialization of the `trainer`

And the same with `self.nth_evaluation`

### DIFF
--- a/trlx/trainer/accelerate_base_trainer.py
+++ b/trlx/trainer/accelerate_base_trainer.py
@@ -129,10 +129,7 @@ class AccelerateRLTrainer(BaseRLTrainer):
                     "Set `tracker` to `None` to disable tracking."
                 )
 
-        self.setup_generate_sweep_kwarg()
         self.nth_evaluation = 0
-
-    def setup_generate_sweep_kwarg(self):
         self.generate_sweep_kwarg = None
         for k, v in self.config.method.gen_kwargs.items():
             if isinstance(v, list):
--- a/trlx/trainer/accelerate_base_trainer.py
+++ b/trlx/trainer/accelerate_base_trainer.py
@@ -130,6 +130,7 @@ class AccelerateRLTrainer(BaseRLTrainer):
                 )
 
         self.generate_sweep_kwarg = None
+        self.nth_evaluation = 0
 
     def setup_model(self):
         """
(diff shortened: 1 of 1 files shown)


reciprocated:
Makes sense! Maybe also let's move the block that initializes `generate_sweep_kwarg` along into trainer's initialization?

mymusise:
> Makes sense! Maybe also let's move the block that initializes `generate_sweep_kwarg` along into trainer's initialization?

Makes sense! The initialization of `generate_sweep_kwarg` depends only on `trainer.config`.

trlx/trainer/accelerate_base_trainer.py
```diff
@@ -129,6 +129,19 @@ def __init__(self, config, **kwargs):  # noqa: C901
                     "Set `tracker` to `None` to disable tracking."
                 )
 
+        self.setup_generate_sweep_kwarg()
+        self.nth_evaluation = 0
+
+    def setup_generate_sweep_kwarg(self):
```
reciprocated, line 135: let's drop this function definition, since it's really single purpose and cannot be used elsewhere, and keep the code block below as is

reciprocated (changes requested):

trlx/trainer/accelerate_base_trainer.py
mymusise: > let's drop this function definition, since it's really single purpose and cannot be used elsewhere, and keep the code block below as is

makes sense

reciprocated (approved):
LGTM!

Issue #459, opened 2023-04-26
OOM issues using Ray+Accelerate for finetuning GPT-J-6B using ILQL


Rohan138:
### 🚀 The feature, motivation, and pitch

Trying to use Ray+Accelerate to finetune GPT-J-6B using ILQL runs into CUDA OOMs.

Code:
```
import json
import os
import sys
import ray
from ray.train.huggingface.accelerate import AccelerateTrainer
from ray.air import ScalingConfig

from datasets import load_dataset

import trlx
from trlx.data.default_configs import (
    ILQLConfig,
    ModelConfig,
    OptimizerConfig,
    SchedulerConfig,
    TokenizerConfig,
    TrainConfig,
    TRLConfig,
)

config = TRLConfig(
    train=TrainConfig(
        seq_length=64,
        batch_size=1,
        epochs=100,
        total_steps=20000,
        checkpoint_interval=10000,
        eval_interval=1000,
        pipeline="PromptPipeline",
        trainer="AccelerateILQLTrainer",
        checkpoint_dir="checkpoints/ilql_hh",
    ),
    model=ModelConfig(model_path="EleutherAI/gpt-j-6B", num_layers_unfrozen=-1),
    tokenizer=TokenizerConfig(tokenizer_path="EleutherAI/gpt-j-6B", truncation_side="left"),
    optimizer=OptimizerConfig(name="adamw", kwargs=dict(lr=1e-6, betas=(0.9, 0.95), eps=1.0e-8, weight_decay=1.0e-6)),
    scheduler=SchedulerConfig(name="cosine_annealing", kwargs=dict(T_max=1000000000, eta_min=1e-6)),
    method=ILQLConfig(
        name="ilqlconfig",
        tau=0.6,
        gamma=0.99,
        cql_scale=0.1,
        awac_scale=1,
        alpha=0.0001,
        beta=0,
        steps_for_target_q_sync=1,
        two_qs=True,
        gen_kwargs=dict(max_new_tokens=56, top_k=20, beta=1, temperature=1.0),
    ),
)

def preprocess(sample):
    sample["prompt_output"] = [
        [sample["prompt"], sample["chosen"]],
        [sample["prompt"], sample["rejected"]],
    ]
    sample["reward"] = [1, -1]
    return sample


def main():
    dataset = load_dataset("Dahoas/static-hh").map(preprocess)
    prompts_outputs = sum(dataset["train"]["prompt_output"], [])
[...]

Issue #372, opened 2023-03-16
Cuda OOM with PPO  on GPT2-medium


OleksandrKorovii:
### 🐛 Describe the bug

I try to train `microsoft/DialoGPT-medium` with trlx. I have 2xRTX3090Ti. No matter what I do, I get `cuda out of memory`. For the reward model, I used fine-tuned my data sts-roberta-base. I tried to decrease `batch_size=2` and left unfreeze the last 2 layers, but it does not help. 
But when I fine-tuned dialoGPT-medium like the casual model on my setup I could put batch_size=10 with seq_len=1024. 

My train pipeline is familiar with examples `ppo-sentiments.py`. I used next following config: 

```
config = TRLConfig(
        train=TrainConfig(
            seq_length=256,
            epochs=5,
            total_steps=10000,
            batch_size=32,
            checkpoint_interval=10000,
            eval_interval=100,
            pipeline="PromptPipeline",
            trainer="AcceleratePPOTrainer",
        ),
        model=ModelConfig(model_path="microsoft/DialoGPT-medium", num_layers_unfrozen=2),
        tokenizer=TokenizerConfig(tokenizer_path="microsoft/DialoGPT-medium", truncation_side="right"),
        optimizer=OptimizerConfig(
            name="adamw", kwargs=dict(lr=1.0e-4, betas=(0.9, 0.95), eps=1.0e-8, weight_decay=1.0e-6)
        ),
        scheduler=SchedulerConfig(name="cosine_annealing", kwargs=dict(T_max=10000, eta_min=1.0e-4)),
        method=PPOConfig(
            name="PPOConfig",
            num_rollouts=128,
            chunk_size=128,
            ppo_epochs=1,
            init_kl_coef=0.05,
            target=6,
            horizon=10000,
            gamma=1,
            lam=0.95,
            cliprange=0.2,
            cliprange_value=0.2,
            vf_coef=1,
            scale_reward="ignored",
            ref_mean=None,
            ref_std=None,
            cliprange_reward=10,
            gen_kwargs=dict(
                max_new_tokens=120,
                top_k=0,
                top_p=1.0,
                do_sample=True,
            ),
        ),
    )

```

Maybe someone can help me with the configuration of the config, and how to run `DialoGPT-medium` with ppo on 2xRTX3090Ti?

### Which trlX version are you using?

`trlx==0.5.0`

### Additional system and package information

[...]

PR #470, opened 2023-05-08
feat(readme): add instructions to avoid OOMs with hyperparameters

maxreciprocate:
This PR adds a small note explaining which config parameters affect the memory usage, while previously it wasn't obvious which should be tweaked and how in order to avoid out of memory issues.

Fix of #372, #459

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,14 +68,28 @@ trainer.generate(**tokenizer('Q: Who rules the world? A:', return_tensors='pt'),
 #### Configure Hyperparameters
 
 ```python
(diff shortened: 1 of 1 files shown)


Dahoas:
@maxreciprocate This is nice to include. It's been sitting around for awhile so wanted to check if you have anything else you want to add? Otherwise I will merge.

Dahoas (approved):

maxreciprocate:
@Dahoas I have nothing else to add here and if it looks good then let's merge!

PR #471, opened 2023-05-09
fix(modeling_ilql): single q-head indexing

maxreciprocate:
This PR fixes an indexing error when using a single q-head instead of two.
To reproduce the error on main: `python examples/randomwalks/ilql_randomwalks.py '{"method": {"two_qs": false}}'`

https://wandb.ai/sorry/trlx/reports/Single-Q-vs-two-Qs-on-randomwalks--Vmlldzo0MzA2NTc2


Dahoas:
Looks good to me. We can merge, unless this was addressed in a more recent pr? @maxreciprocate

maxreciprocate:
@Dahoas It was not addressed in any of recent prs

Dahoas:
Let's merge then

Issue #473, opened 2023-05-10
AttributeError: 'NoneType' object has no attribute 'items'


zwhe99:
### 🐛 Describe the bug

I tried to follow the example given in the readme, but unfortunately, I couldn't succeed.

```python
from trlx import trlx
trainer = trlx.train('EleutherAI/gpt-neo-125m', samples=['dolphins', 'geese'], rewards=[1.0, 100.0])
``` 
```shell
[RANK 0] Collecting rollouts
[RANK 0] Logging sample example
                  Sample Example                  
┏━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━┓
┃ Prompt        ┃ Response              ┃ Reward ┃
┡━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━┩
│ <|endoftext|> │ dolphins<|endoftext|> │ 1.0    │
└───────────────┴───────────────────────┴────────┘
[RANK 0] Logging experience string statistics
   Experience String Stats (mean ∈ [min, max])   
┏━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━┓
┃ Prompt Length ┃ Output Length ┃ Sample Length ┃
┡━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━┩
│ 1.00 ∈ [1, 1] │ 3.00 ∈ [3, 3] │ 4.00 ∈ [4, 4] │
└───────────────┴───────────────┴───────────────┘
[RANK 0] Starting training
[RANK 0] Evaluating model
[generation sweep 0/1 | eval batch 0/1]:   0%|                                                                                       | 0/1 [00:00<?, ?it/s]You're using a GPT2TokenizerFast tokenizer. Please note that with a fast tokenizer, using the `__call__` method is faster than using a method to encode the text followed by a call to the `pad` method to get a padded encoding.
[generation sweep 1/1 | eval batch 1/1]: 100%|███████████████████████████████████████████████████████████████████████████████| 1/1 [00:05<00:00,  5.32s/it]
[RANK 0] Summarizing evaluation
                                                                       Evaluation #0                                                                       
[...]

PR #475, opened 2023-05-10
fix(offline_pipeline): force `drop_last` only for distributed

reciprocated:
This PR keeps `drop_last` policy for dataloaders only for multiprocess training to avoid duplication, but keeps last batches for single processes to avoid losing data. It also skips warning messages that `MiniBatchIterator` could throw even though minibatching is not actived, for example when running `python examples/randomwalks/ppo_randomwalks.py` on the current `main`

Fix of #473


Dahoas (approved):
LGTM, nice fix

Issue #434, opened 2023-04-13
Migrate to `peft` from `opendelta` for parameter efficient tuning methods


jon-tow:
### 🚀 The feature, motivation, and pitch

Let's migrate to [`peft`](https://github.com/huggingface/peft).

Citing @ethankim00's concerns with `opendelta`:
- `opendelta` import fails due to an unnecessary `turtle` package import. Even if pip installed, users may be required to have `sudo` privileges to install the corresponding base graphics package https://github.com/thunlp/OpenDelta/issues/47
- Doesn’t seem to work with DeepSpeed ZeRO 3
- Additional inference overhead from not merging in the LoRA adapters layers
- Incompatibility with int8 training
- Less actively maintained than the `peft` library, which has been growing rapidly
- Sharing adapter weights on the HuggingFace Hub is less convenient with `opendelta`

jon-tow:
Assignee: @glerzing will be taking a go at this :)

loganlebanoff:
I'm curious as to the status of this issue/PR

glerzing:
I'm developing automated tests for it, there should be a PR soon.

akk-123:
@glerzing look forward it

akk-123:
@glerzing when will have a pr?

glerzing:
It should be ready for PR tomorrow, sorry for the wait.

Issue #320, opened 2023-02-21
Collapse reference-learner hydra heads when using LoRa


jon-tow:
### 🚀 The feature, motivation, and pitch

With additive (delta-style) parameter-efficient tuning methods such as [LoRa](https://arxiv.org/abs/2106.09685), we should be able to make a slightly more mem-efficient hydra architecture by using a single block that does ~`frozen_head + tunable_weights` for the learner/policy head's fwd-pass and simply `frozen_head` for the reference, instead of maintaining 2x heads.

CC @LouisCastricato and @cat-state for pointing this out

LouisCastricato:
Haha, I was not aware that Aman proposed the same thing.

LouisCastricato:
I am not sure anyone started. cc @jon-tow

jon-tow:
Not yet. @glerzing is looking into `peft` migration, which should make this very simple as the package provides a context manager to disable adapters https://github.com/huggingface/peft/blob/34027fe813756897767b9a6f19ae7f1c4c7b418c/src/peft/peft_model.py#L290 that we can use over reference model calls.

Opdoop:
Cool!!!  Looking forward to this update. 👍

glerzing:
Sorry to make you wait, but it should take a few weeks to get this done. As a very rough estimate, I would say that I may push a tested solution around the 10th may. But I'm new here so I don't know how much time would then happen before a new release version.

PR #477, opened 2023-05-13
peft to opendelta migration (#434) + memory optimization (#320)

glerzing:
It was simpler to solve both issues in the same PR.

peft is not completely stable, there are still a few bugs, especially on seq2seq (e.g. flan-t5-small does not work). LORA looks pretty stable though. I mostly relied on unit tests, but I will keep testing a few things in the coming days. And I will maybe add in a next PR a use-case example if you think that's useful. I removed "ref_model" because it didn't seem to be actually used.

Feel free to ask questions or to make remarks. For example on the coding style, I don't know if there are some guidelines. Also, if you want the automated tests to be faster, I can reduce the number of models or configs to test.


glerzing:
Some things I could do next, tell me if you agree :
- See if I can make it possible to use multiple methods (e.g. prefix tuning + Lora)
- Make it possible to retrain a peft fine-tuned model (currently trlx can create but can't load a peft model)

glerzing:
Is it ok if we save only the peft adapter and not the value head ?

LouisCastricato:
no there are many situations you would need to save the value head. also yes we should have loading, especially for resuming from checkpoints.

I am going to mark this as draft as it seems to be not ready for review. I do think in the interim you should resolve the loading/saving issues and add more unit tests.

glerzing:
> no there are many situations you would need to save the value head.

But then should we save the whole model instead of just the peft adapter ? Or let a boolean option for whether to save the whole model or just the adapter ? Or make some sophisticated code to save just the peft adapter and the value head ? (I'm not sure how important it is to limit the memory size.)

PR #487, opened 2023-05-24
Add Stable Vicuna Training

PhungVanDuy:


examples/stable_vicuna/rl_training.py
```diff
@@ -0,0 +1,148 @@
...
+
+                with torch.no_grad():
+                    scores = rm_model(**batch)[0].squeeze(-1).cpu()
+                all_scores.append(scores)
+            scores = torch.hstack(all_scores)
+            return scores
+
+        def reward_fn(samples, prompts, original_output, **kwargs):
+            samples = [s[s.find("<|prompter|>") :] for s in samples]
+            prompts = [p[p.find("<|prompter|>") :] for p in prompts]
```
maxreciprocate, line 118: Why this has to be done here? Is it to strip `<|system|>` from some of the prompts?

Issue #483, opened 2023-05-22
RuntimeError using Accelerate + Zero-3 to launch `ppo_sentiments_llama.py` (uninitialized LayerNorm weight in Hydra head?)


mbalesni:
### 🐛 Describe the bug

When I try fine-tuning LLaMA using Accelerate with DeepSpeed Zero-3 with several GPUs, the training crashes when trying to use the Hydra head due to what looks like an uninitialized LayerNorm weight. 

* I want to use a distributed setup to train bigger LLaMAs than 7b.
* The error in this issue is different from #461 (which I also can replicate, but with non-LLaMA models)
* Note that I use a [fork of `trlx`](https://github.com/meg-tong/trlx) (forked a week ago) but it only changes the example-relevant code to load LLaMA locally and adds BOS tokens to the tokenized inputs. I tested it to work correctly on 1 GPU (without `accelerate`), and I believe you can ignore this nuance and treat it as me using the original `trlx` from the main branch.

1. Command:

```bash
accelerate launch trlx/examples/ppo_sentiments_llama.py
```

2. The error trace:

```
/data/mykyta_baliesnyi/situational-awareness/trlx/examples/ppo_sentiments_llama.py:122 in        │
│ <module>                                                                                         │
│                                                                                                  │
│   119 │   if args.debug:                                                                         │
│   120 │   │   attach_debugger()                                                                  │
│   121 │                                                                                          │
│ ❱ 122 │   main(args)                                                                             │
│   123                                                                                            │
│                                                                                                  │
│ /data/mykyta_baliesnyi/situational-awareness/trlx/examples/ppo_sentiments_llama.py:105 in main   │
│                                                                                                  │
│   102 │   imdb = load_dataset("imdb", split="train+test")                                        │
│   103 │   prompts = [" ".join(review.split()[:4]) for review in imdb["text"]]                    │
│   104 │                                                                                          │
│ ❱ 105 │   trlx.train(                                                                            │
│   106 │   │   reward_fn=reward_fn,                                                               │
│   107 │   │   prompts=prompts,                                                                   │
│   108 │   │   eval_prompts=["I don't know much about Hungarian underground"] * 64,               │
│                                                                                                  │
│ /data/mykyta_baliesnyi/situational-awareness/trlx/trlx/trlx.py:105 in train                      │
│                                                                                                  │
│   102 │   │   if eval_prompts is None:                                                           │
│   103 │   │   │   eval_prompts = prompts[:batch_size]                                            │
[...]

Issue #461, opened 2023-04-28
RunTimeError using Accelerate + Zero Stage 3 to launch ppo_sentiments.py


alex-athanassakos:
### 🐛 Describe the bug

Hi!

I have been trying to use Accelerate with Deep Speed to launch my TRLX scripts and have been running into this error in several instances:

```╭───────────────────── Traceback (most recent call last) ─────────────────────╮
│ /home/ubuntu/Codes/resolver-ai/prompt_resp_gen_train/RL/ppo_sentiments.py:5 │
│ 7 in <module>                                                               │
│                                                                             │
│   54                                                                        │
│   55 if __name__ == "__main__":                                             │
│   56 │   hparams = {} if len(sys.argv) == 1 else json.loads(sys.argv[1])    │
│ ❱ 57 │   main(hparams)                                                      │
│   58                                                                        │
│                                                                             │
│ /home/ubuntu/Codes/resolver-ai/prompt_resp_gen_train/RL/ppo_sentiments.py:4 │
│ 7 in main                                                                   │
│                                                                             │
│   44 │   imdb = load_dataset("imdb", split="train+test")                    │
│   45 │   prompts = [" ".join(review.split()[:4]) for review in imdb["text"] │
│   46 │                                                                      │
│ ❱ 47 │   trlx.train(                                                        │
│   48 │   │   reward_fn=reward_fn,                                           │
│   49 │   │   prompts=prompts,                                               │
│   50 │   │   eval_prompts=["I don't know much about Hungarian underground"] │
│                                                                             │
│ /home/ubuntu/Codes/resolver-ai/prompt_resp_gen_train/RL/venv/src/trlx/trlx/ │
│ trlx.py:101 in train                                                        │
│                                                                             │
│    98 │   │   if eval_prompts is None:                                      │
│    99 │   │   │   eval_prompts = prompts[:batch_size]                       │
│   100 │   │                                                                 │
│ ❱ 101 │   │   trainer.make_experience(config.method.num_rollouts)           │
│   102 │                                                                     │
│   103 │   # Offline training from the collected samples (e.g. SFT, ILQL)    │
│   104 │   elif samples:                                                     │
│                                                                             │
│ /home/ubuntu/Codes/resolver-ai/prompt_resp_gen_train/RL/venv/src/trlx/trlx/ │
│ trainer/accelerate_ppo_trainer.py:409 in make_experience                    │
│                                                                             │
│   406 │   │   │   │   │   )                                                 │
[...]

PR #489, opened 2023-05-26
fix(modeling_ppo): load reference head under zero3

maxreciprocate:
Fix of #461 #483

Under zero3 the reference head will be kept unsharded. I couldn't get sharding to work because deepspeed expects sharded parameters to have at least one backward pass, while there is no backward pass for the reference head.

Check:
```bash
accelerate launch --config_file configs/accelerate/zero3.yaml examples/ppo_sentiments.py
```

https://wandb.ai/sorry/trlx-references/reports/fix-ppo-zero3-v-main--Vmlldzo0NjY3MDYy

### DIFF
--- a/trlx/models/modeling_ppo.py
+++ b/trlx/models/modeling_ppo.py
@@ -1,13 +1,12 @@
 import gc
-import deepspeed
 import inspect
 from copy import deepcopy
 from dataclasses import dataclass
 from typing import List, Optional, Tuple, Union
 
+import deepspeed
 import numpy as np
 import torch
-import torch.nn as nn
 import transformers
 from torchtyping import TensorType
 from transformers.modeling_outputs import ModelOutput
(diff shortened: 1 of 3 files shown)


nikebless:
I don't think unsharding is necessary for zero3 in general. It is only necessary for `zero3_init_flag: true`.

I ran your changes locally, undoing the `modeling_ppo.py` changes and updating `configs/accelerate/zero3.yaml` to `zero3_init_flag: false`, and the training worked (side note: passing `--zero3_init_flag false` as an argument doesn't override the config file)

trlx/models/modeling_ppo.py
```diff
@@ -410,10 +410,18 @@ def __init__(
...
-        decoder_blocks = deepcopy(hf_get_decoder_blocks(base_model))
-        self.decoder_blocks = nn.ModuleList(list(decoder_blocks)[-num_layers_unfrozen:])
-        self.final_norm = deepcopy(hf_get_decoder_final_norm(base_model))
-        self.lm_head = deepcopy(hf_get_lm_head(base_model))
+
+        decoder_blocks = hf_get_decoder_blocks(base_model)[-num_layers_unfrozen:]
+        final_norm = hf_get_decoder_final_norm(base_model)
+        lm_head = hf_get_lm_head(base_model)
+
+        with deepspeed.zero.GatheredParameters(
```
Dahoas, line 418: Should this be dependent on whether we are using zero3 (zero3_init)?

Dahoas: @maxreciprocate

maxreciprocate: @Dahoas Not necessary, this context would be a noop when zero3 is not enabled and would still work correctly, see comparisons runs https://wandb.ai/sorry/trlx-references/reports/fix-ppo-zero3-v-main--Vmlldzo0NjY3MDYy

Dahoas (approved):
Nice, if you have no other changes let's merge this in.

PR #496, opened 2023-06-03
fix type hint in PromptPipeline.__init__

### LINKED ISSUE #494: Type hint doesn't match documentation for prompts arg of PromptPipeline.__init__
https://github.com/CarperAI/trlx/blob/0dce99d96b7d70b6a9114129d8e38bf6c80eb653/trlx/pipeline/offline_pipeline.py#L108-L116

https://github.com/CarperAI/trlx/blob/0dce99d96b7d70b6a9114129d8e38bf6c80eb653/trlx/pipeline/offline_pipeline.py#L126

g-simmons:
Fixes #494

### DIFF
--- a/trlx/pipeline/offline_pipeline.py
+++ b/trlx/pipeline/offline_pipeline.py
@@ -123,7 +123,7 @@ class PromptPipeline(BasePipeline):
 
     def __init__(
         self,
-        prompts: Union[Dict[str, Any], List[str]],
+        prompts: Union[List[Dict[str, Any]], List[str]],
         max_prompt_length: int,
         tokenizer: PreTrainedTokenizer,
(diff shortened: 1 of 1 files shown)


LouisCastricato:
Ah I assumed you'd change the doc string. Did you verify that this change doesn't break anything?

g-simmons:
Why I thought this change was correct: 
- `eval_prompts` is documented to accept `List[Dict[str,Any]]` in `train` fn of `trlx.py`
- it seems like the first element of prompts is expected to sometimes be a dict in `PromptPipeline.__init__`.

cat-state (approved):
LGTM! Thanks @g-simmons

PR #510, opened 2023-06-23
fix(ppo_trainer): default gen kwargs

maxreciprocate:
This PR 
- forces `use_cache` in `gen_kwargs` to be true by default
- replaces `batch_size` to `chunk_size` in evaluation pipeline, mirroring `make_experience` pipeline usage
- removes enforcement on padding token to be the same as endoftext with decoder models

https://wandb.ai/sorry/trlx-references/reports/fix-default-gen_kwargs-v-main--Vmlldzo0NzIwNzc2
(note that pythia run is absent because there are only 40gb gpus available currently)


PhungVanDuy (approved):
sgtm!

PR #514, opened 2023-06-26
Dense reward carper (Fine grained feedback)

Dahoas:


trlx/trainer/accelerate_ppo_trainer.py
```diff
@@ -297,24 +298,39 @@ def make_experience(self, num_rollouts: int = 1024, iter_count: int = 0):  # noq
...
+                    **metadata,
                 )
+                all_scores = [
+                    torch.tensor(score, dtype=torch.float, device=device).view(
+                        -1,
+                    )
+                    for score in all_scores
+                ]
+                # Pad 0 reward on the ends
+                all_scores = pad_sequence(all_scores, batch_first=True, padding_value=-1)
```
maxreciprocate, line 317: Is this padding done under assumption that scores are strictly possitive? Otherwise `padding_value=-1` would be incorrect here, and later when computing the `scores_mask`
```suggestion
                all_scores = pad_sequence(all_scores, batch_first=True, padding_value=-np.inf)
```

trlx/trainer/accelerate_ppo_trainer.py
```diff
@@ -342,8 +358,10 @@ def make_experience(self, num_rollouts: int = 1024, iter_count: int = 0):  # noq
 
             # store statistics of the initial rollout as reference
             if self.ref_mean is None:
-                self.ref_mean, self.ref_std = scores.mean(), scores.std()
-            all_scores_mean, all_scores_std = self.running_moments.update(scores)
+                self.ref_mean, self.ref_std = (scores * scores_mask).sum(dim=1).mean(), (scores * scores_mask).sum(
```
maxreciprocate, line 361: Do you want to take here sum across tokens, or just mean i.e. `.mean(dim=1)`? Although you're not using any type of score normalization in your example, for the case when it's used (`scores /= self.ref_std`) taking mean instead of sum to compute `self.ref_std` should maybe work better here, if I understand correctly

maxreciprocate: Nit: you can also precompute `scores * scores_mask` and pass it later into `self.running_moments.update`

Dahoas: I also wasn't sure whether it's better to normalize by returns or rewards. This [paper](https://arxiv.org/pdf/2105.05347.pdf) suggests normalizing by returns is better (at least when using multiple reward scales) so I went with that. Also the reward mean estimate is sensitive to the trajectory length, which I think we want to avoid.

Dahoas: > Nit: you can also precompute `scores * scores_mask` and pass it later into `self.running_moments.update`

Good point, I will change this.

Dahoas:
How do I fix this build error during the testing?

Glavin001:
Test errors:

```
=========================== short test summary info ============================
FAILED tests/test_trainers.py::TestAccelerateBaseTrainer::test_save_checkpoint - RuntimeError: mean(): could not infer output dtype. Input dtype must be either a floating point or complex dtype. Got: Long
FAILED tests/test_trainers.py::TestAccelerateBaseTrainer::test_save_lora_checkpoint - RuntimeError: mean(): could not infer output dtype. Input dtype must be either a floating point or complex dtype. Got: Long
======= 2 failed, 70 passed, 3 skipped, 22 warnings in 803.24s (0:13:23) =======
Error: Process completed with exit code 1.
```

A bit more detail:

```
tests/test_trainers.py::TestAccelerateBaseTrainer::test_save_checkpoint FAILED [ 82%]
tests/test_trainers.py::TestAccelerateBaseTrainer::test_save_lora_checkpoint FAILED [ 84%]
...
________________ TestAccelerateBaseTrainer.test_save_checkpoint ________________

self = <tests.test_trainers.TestAccelerateBaseTrainer testMethod=test_save_checkpoint>

    def test_save_checkpoint(self):
        with tempfile.TemporaryDirectory() as tmpdir:
            config = self.get_default_config()
            config.train.checkpoint_dir = tmpdir
    
            trainer = self.get_trainer(config)
>           trainer.learn()

tests/test_trainers.py:126: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
trlx/trainer/accelerate_base_trainer.py:537: in learn
    results = self.evaluate()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <trlx.trainer.accelerate_ppo_trainer.AcceleratePPOTrainer object at 0x7f7eaedd5f40>

   def evaluate(self):  # noqa: C901
...
>                   mean_reward = rewards.mean().item()
E                   RuntimeError: mean(): could not infer output dtype. Input dtype must be either a floating point or complex dtype. Got: Long
```

Therefore, applicable code is here:

trlx/trainer/accelerate_base_trainer.py
```diff
@@ -425,10 +425,19 @@ def evaluate(self):  # noqa: C901
...
+                        outputs=str_outputs,
+                        model_tok=self.tokenizer,
+                        **metadata,
                     )
+                    if type(rewards[0]) is torch.Tensor:
+                        rewards = torch.tensor([reward.sum().item() for reward in rewards], dtype=float)
+                    elif type(rewards[0]) is list:
+                        rewards = torch.tensor([sum(reward) for reward in rewards])
+                    else:
+                        rewards = torch.tensor(rewards)
```
Glavin001, line 440: Looking at the args into the `reward_fn`, changes highlighted:

```diff
  samples=str_samples,
  prompts=str_prompts,
  outputs=str_outputs,
+ model_tok=self.tokenizer,
  **metadata,
```

I'll assume that's intentional.

Looking at the args into the different calls to `torch.tensor`:

```diff
# Assume
raw_rewards = self.reward_fn(...)

# Convert raw to tensor rewards
rewards = torch.tensor(
                        raw_rewards,
-                        dtype=float, # Missing this?
                  )
```

Missing `dtype=float` on lines:
- ✅  436
- ❌ 438
- ❌ 440

Therefore, proposed fix:

```suggestion
                    raw_rewards = self.reward_fn(
                        samples=str_samples,
                        prompts=str_prompts,
                        outputs=str_outputs,
                        model_tok=self.tokenizer,
                        **metadata,
                    )
                    if type(rewards[0]) is torch.Tensor:
                        rewards = torch.tensor([reward.sum().item() for reward in rewards], dtype=float)
                    elif type(rewards[0]) is list:
                        rewards = torch.tensor([sum(reward) for reward in rewards], dtype=float)
                    else:
                        rewards = torch.tensor(rewards, dtype=float)
```


**Disclaimer**: I suck/new at Python (JavaScript dev), never used trlx (want to), so please ignore if these differences were intentional ❤️ .

trlx/trainer/accelerate_ppo_trainer.py
```diff
@@ -297,24 +299,39 @@ def make_experience(self, num_rollouts: int = 1024, iter_count: int = 0):  # noq
...
                 )
+                all_scores = [
+                    torch.tensor(score, dtype=torch.float, device=device).view(
+                        -1,
+                    )
+                    for score in all_scores
+                ]
+                # Pad 0 reward on the ends
+                all_scores = pad_sequence(all_scores, batch_first=True, padding_value=-np.inf)
+                max_len = torch.tensor(all_scores.shape[1], dtype=torch.long, device=device)
```
Glavin001, line 319: Given the error message:

> RuntimeError: mean(): could not infer output dtype. Input dtype must be either a floating point or complex dtype. Got: Long

```suggestion
                max_len = torch.tensor(all_scores.shape[1], dtype=torch.float, device=device)
```

trlx/trainer/accelerate_base_trainer.py
Glavin001: PR applying this fix: https://github.com/CarperAI/trlx/pull/520

trlx/trainer/accelerate_ppo_trainer.py
maxreciprocate: No, `max_len` is used as an argument for `reshape`, therefore it has to be long

Glavin001:
What're next steps to getting this merged? 😃  Really excited to try it out! 🎉 Thank you to all contributors!

trlx/trainer/accelerate_ppo_trainer.py
```diff
@@ -452,7 +472,17 @@ def make_experience(self, num_rollouts: int = 1024, iter_count: int = 0):  # noq
...
+                if scores.shape[1] == 1:
+                    # NOTE: Final reward given at EOS token following HHH practice
+                    rewards[-1] += scores[sample_idx][0].cpu()
+                else:
+                    score = scores[sample_idx]
+                    score_right_padding = torch.sum(scores_mask[sample_idx])
+                    score = score[:score_right_padding].cpu()
+                    p_score = torch.zeros_like(rewards)
+                    p_score[: score.shape[0]] += score
+                    rewards += p_score
```
maxreciprocate, line 485: Nit: could you drop a comment here for what a `p_score` is, and why these computations are needed

trlx/trainer/accelerate_ppo_trainer.py
```diff
@@ -297,24 +299,39 @@ def make_experience(self, num_rollouts: int = 1024, iter_count: int = 0):  # noq
...
-                    ),
-                    dtype=torch.float,
-                    device=device,
+                # reward_fn should return list of rewards at each token per sample
+                # NOTE: all_scores[0][i] is the reward due to token (action) i in prompt + response (b/c of how kl is computed)
+                all_scores = self.reward_fn(
+                    samples=all_str_samples,
+                    prompts=all_str_prompts,
+                    outputs=all_str_outputs,
+                    model_tok=self.tokenizer,
```
maxreciprocate, line 308: Nit: maybe rename this argument from `model_tok` to just `tokenizer`, since if it's used, the distinction disappears

trlx/trainer/accelerate_base_trainer.py
```diff
@@ -427,10 +427,19 @@ def evaluate(self):  # noqa: C901
...
-                        self.reward_fn(samples=str_samples, prompts=str_prompts, outputs=str_outputs, **metadata),
-                        dtype=float,
+                    rewards = self.reward_fn(
+                        samples=str_samples,
+                        prompts=str_prompts,
+                        outputs=str_outputs,
+                        model_tok=self.tokenizer,
+                        **metadata,
                     )
+                    if type(rewards[0]) is torch.Tensor:
```
maxreciprocate, line 437: Nit:
```suggestion
                    if isinstance(rewards[0], torch.Tensor)
```

maxreciprocate (changes requested):
No difference against main https://wandb.ai/sorry/trlx-references/reports/dense-reward-carper-v-main--Vmlldzo0OTAxNzUx

trlx/trainer/accelerate_ppo_trainer.py
Dahoas: This is to guard against sizing errors due to the tokenisation issues we discussed today. It is fixed in a later pr.

maxreciprocate (approved):
Happy to merge this, as I see there's some overlap with your most recent PRs, although we probably should document this feature, because if not for this newly added example, it would stay rather hidden. What do you think?

Dahoas:
Yes we should definitely document. I can do that once all the recent prs get merged.

trlx/trainer/accelerate_base_trainer.py
```diff
@@ -427,10 +427,19 @@ def evaluate(self):  # noqa: C901
...
                 if self.reward_fn:
                     logger.info("Computing rewards")
-                    rewards = torch.tensor(
-                        self.reward_fn(samples=str_samples, prompts=str_prompts, outputs=str_outputs, **metadata),
-                        dtype=float,
+                    rewards = self.reward_fn(
+                        samples=str_samples,
+                        prompts=str_prompts,
+                        outputs=str_outputs,
+                        model_tok=self.tokenizer,
```
Glavin001, line 434: Missed a spot. Should have been renamed from `model_tok` to `tokenizer`.

image

Dahoas: Thanks for reporting this!

PR #515, opened 2023-07-05
fix peft config cause tensorboard type error

congchan:
peft config is not supported by tensorboard add hyperparameters. So we parse it to string.


maxreciprocate (approved):
LGTM! Thanks, Cong!

PR #516, opened 2023-07-05
fix(modeling_base): handle parent directory for peft head save

congchan:
Parent directory not found error occurs in ppo training, because `torch.save` won't handle the save path for us(https://github.com/pytorch/pytorch/issues/71389).


maxreciprocate:
LGTM!

maxreciprocate (approved):
LGTM!

PR #519, opened 2023-07-08
Add Gradient accumulation steps to Accelerate Trainer

uSaiPrashanth:
This PR mostly is with one-off nitpicks that I changed while performing demo runs with trlx

Changes made:
- Added gradient accumulation steps to accelerate trainer
- Changed wandb run name to not have "unknown" branch name when installed via pip


allzero-kwon:
Thank you! This is helpful for me. :)

PR #522, opened 2023-07-13
Fix ordering of ppo epoch iteration

RobertKirk:
Suppose you have 256 rollouts, and you batch them into batches B1, B2, B3, B4 (each of size 64). The order of gradient updates (assuming 3 ppo_epochs) is:

`trlx: B1 B1 B1 B2 B2 B2 B3 B3 B3 B4 B4 B4`

However, what we should actually be doing (and what alpaca-farm and other rlhf implementations, and standard implementations of PPO do), is

`optimal: B1 B2 B3 B4 B1' B2' B3' B4' B1* B2* B3* B4*`

This change reorders the learning to make the code use the `optimal` ordering above. It also renames n_updates_per_batch to n_inner_epochs as that's a more accurate description (especially now), adjusts forward_time and backward_time to not type-error, and renames mbs and mb to minibatch and microbatch (as that's what they are).

### DIFF
--- a/trlx/trainer/accelerate_base_trainer.py
+++ b/trlx/trainer/accelerate_base_trainer.py
@@ -645,11 +645,10 @@ class AccelerateRLTrainer(BaseRLTrainer):
             self.post_epoch_callback()
         tbar.close()
 
-    def create_train_dataloader(self, shuffle=True, accelerate_prepare=True):
-        dataloader = self.store.create_loader(self.config.train.batch_size, shuffle=shuffle)
-        if accelerate_prepare:
-            dataloader = self.accelerator.prepare_dataloader(dataloader)
-        return dataloader
+    @abstractmethod
+    def create_train_dataloader(self):
+        """Returns a new dataloader for training."""
+        pass
 
     @abstractmethod
     def get_arch(self, config: TRLConfig):
--- a/trlx/trainer/accelerate_base_trainer.py
+++ b/trlx/trainer/accelerate_base_trainer.py
@@ -542,21 +542,21 @@ class AccelerateRLTrainer(BaseRLTrainer):
 
         # For each epoch
         for _ in range(self.config.train.epochs):
-            # For each batch
-            for mbs in MiniBatchIterator(self.train_dataloader, self.mb_size, self.num_mb):
-                # For each update per batch
-                for _ in range(self.n_updates_per_batch):
-                    # Note that whereas standard policy gradient methods perform one
-                    # gradient update per batch, PPO for example commonly performs
-                    # multiple gradient updates on the same batch of data.
-                    # https://arxiv.org/pdf/1707.06347.pdf
-                    forward_time = 0
-                    backward_time = 0
+            # For each ppo epoch
+            for _ in range(self.n_inner_epochs):
+                # Note that whereas standard policy gradient methods perform one
+                # gradient update per batch, PPO for example commonly performs
+                # multiple epochs of gradient updates on the same batch of data.
+                # https://arxiv.org/pdf/1707.06347.pdf
+                # For each batch
+                for minibatch in MiniBatchIterator(list(self.train_dataloader), self.mb_size, self.num_mb):
+                    forward_time = 0.0
+                    backward_time = 0.0
                     stats_accum = []
-                    for mb in mbs:
+                    for microbatch in minibatch:
                         with self._accumulate():
                             forward_time -= time()
-                            loss, stats = self.loss(mb)
+                            loss, stats = self.loss(microbatch)
                             forward_time += time()
                             backward_time -= time()
                             self.accelerator.backward(loss)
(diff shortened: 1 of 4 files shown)


trlx/trainer/accelerate_base_trainer.py
```diff
@@ -542,21 +542,21 @@ def learn(self):  # noqa: C901
...
-                    forward_time = 0
-                    backward_time = 0
+            # For each ppo epoch
+            for _ in range(self.n_inner_epochs):
+                # Note that whereas standard policy gradient methods perform one
+                # gradient update per batch, PPO for example commonly performs
+                # multiple epochs of gradient updates on the same batch of data.
+                # https://arxiv.org/pdf/1707.06347.pdf
+                # For each batch
+                for minibatch in MiniBatchIterator(list(self.train_dataloader), self.mb_size, self.num_mb):
```
maxreciprocate, line 552: Nitpick: is there a reason it has to be changed to `list(self.train_dataloader)`?

maxreciprocate (commented):
Splendid! Thank you, for this implementation detail has been left out. Here are comparisions between runs using this PR and the main branch https://wandb.ai/sorry/trlx-references/reports/main-v-main--Vmlldzo0ODc2MTAx (forgive the ambiguity with naming branches `main` and `main`, but respective top runs in wandb are made on this PR). Difference isn't too notable on most of these, however there is a difference for ILQL runs, but since `self.n_inner_epochs=1` for it, this change shouldn't bear an effect there I suppose

RobertKirk:
Maybe I'm reading this wrong, but these results also imply that there is a fair difference (although I can't tell which run is this PR and which is the current main).

image

image

trlx/trainer/accelerate_base_trainer.py
maxreciprocate: Of course, that makes sense! Also dumb question: will `shuffle=True` in `self.train_dataloader` produce optimal behaviour or is it different?

maxreciprocate:
Yeah you're right, that's what I meant by "on most of these", but also I rather meant that there are no obvious deteriorations, it was not in demeaning way towards this change, since it's obviously correct and in line with other implementations

(Top ones are from this PR, terribly sorry for the confusion)

trlx/trainer/accelerate_base_trainer.py
RobertKirk: I've added a commit that abstracts the training dataloader creation and then reinstantiates it at every ppo epoch. It shoudn't have changed the logic for any of the trainers apart from the accelerate_ppo_trainer.

Dahoas:
@maxreciprocate Do we want to merge this now?

trlx/trainer/accelerate_ilql_trainer.py
```diff
@@ -158,18 +158,20 @@ def loss(self, batch: Union[ILQLBatch, ILQLSeq2SeqBatch]):
 
         return self.ilql.loss((logits, (qs, target_qs, vs)), batch)
 
+    def create_train_dataloader(self):
+        return self.accelerator.prepare(self.store.create_loader(self.config.train.batch_size))
```
Dahoas, line 162: It looks like `create_loader` is missing an argument for `shuffle`.

maxreciprocate (commented):
Although while there isn't difference for SFT runs, there is a difference for ILQL runs, however I can't figure out why exactly, since data ordering stays unchanged, as it should 

https://wandb.ai/sorry/trlx-references/reports/main-v-main--Vmlldzo0OTQ2MzEw

trlx/trainer/accelerate_base_trainer.py
```diff
@@ -633,6 +636,12 @@ def learn(self):  # noqa: C901
             self.post_epoch_callback()
         tbar.close()
 
+    def create_train_dataloader(self, shuffle=True, accelerate_prepare=True):
+        dataloader = self.store.create_loader(self.config.train.batch_size, shuffle=shuffle)
+        if accelerate_prepare:
+            dataloader = self.accelerator.prepare_dataloader(dataloader)
```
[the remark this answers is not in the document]
RobertKirk: We could make this an abstract function or a not-implemented function if it's ok for `AccelerateRLTrainer` to be an abstract class or require subclassing, but I didn't know whether that was the case, so I put a standard implementation here instead.

RobertKirk:
It could be seeded differently as we pass the dataloaders to accelerate prepare in a different order than we did before? Or there's uncontrolled non-determinism that the seed isn't setting?

trlx/trainer/accelerate_base_trainer.py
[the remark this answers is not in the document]
maxreciprocate: Yes, let's make it a abstract method just like the ones below it, with only difference between different implementations is that we either shard or not shard dataloader, correct?

maxreciprocate:
Probably, that's most likely the reason, since I've check the order in which tensors come for ILQL and it hasn't changed

trlx/trainer/accelerate_base_trainer.py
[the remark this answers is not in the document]
RobertKirk: The difference is whether the dataloader is prepared by accelerate and whether you pass shuffle=True - ppo doesn't prepare and passess shuffle=True, ILQL and SFT do prepare and don't pass shuffle=True. I've made the change now.

maxreciprocate (approved):
Excellent fix, greatly appreciated @RobertKirk!

PR #529, opened 2023-07-18
Dist ref kl

Dahoas:
Implementing `ref_model` as an additional reward component


trlx/trainer/accelerate_ppo_trainer.py
PhungVanDuy, line 1: in line 472 sample_outputs not assigned yet, seems `sample_outputs = pad_sequence(tok_outputs)`

PR #546, opened 2023-08-21
mosaicml/mpt support

50m-regent:
Since the ModuleList in mpt models from mosaicml is named 'transformer.blocks', its unloadable with current trlx. Simply adding the module name, it was able to train the model :)


maxreciprocate (approved):
Thanks!

PR #547, opened 2023-08-21
chore(requirements.txt): update everything to the latest

maxreciprocate:


Dahoas:
Looks like we'll have to update the minimum python requirement to 3.9 if we want the newest [numpy](https://github.com/numpy/numpy/releases/tag/v1.25.0).

Dahoas (approved):

PR #555, opened 2023-09-04
Inference pipeline

Dahoas:
Implementation of multi-generation RL in trlX

Suggested (but optional) external inference pipeline wrapper can be found[ here](https://github.com/CarperAI/autocrit/pull/16)


Dahoas:
Depends on #529

PR #568, opened 2023-10-17
support extra model and tokenizer configs during loading by from_pretrained in accelerate trainer

Jingru:
Currently, no extra configs (args in `from_pretrained` function, like `trust_remote_code`) are supported in accelerate trainer, which makes SFT/RLHF not impossible for models like baichuan/internlm etc.

So, extra model/tokenizer configs should be customizable.


maxreciprocate (approved):
LGTM! Thanks for this worthy addition!

`internlm/internlm-7b` with `trust_remote_code=True`: https://wandb.ai/sorry/trlx/runs/0metzq47

